PR #62, opened 2026-04-18
Fix #54: Question about input handling during agent runs — relates to...

### LINKED ISSUE #54: Question about input handling during agent runs — relates to "Human in the Loop"
Hi team 👋
First of all, thanks for the amazing work on Decepticon — really enjoying exploring it.
I wanted to ask about something I noticed in the CLI and check whether it's intended behavior or something that might be improved.
What I'm seeing
While agents are running and streaming output, the CLI input seems to be unavailable — I can't type or send anything to the agent until the current run finishes.
What I was expecting
Reading the [Core Philosophy → Human in the Loop](https://docs.decepticon.red/en/vision/core-philosophy#human-in-the-loop) section, it sounds like the operator should be able to stay in the loop during a run:

Real-Time Monitoring: operators can watch the agent's actions as they happen
Intervention at Any Time: the operator can pause, redirect, or override the agent's decisions at any moment
Feedback-Driven Learning: human feedback shapes the agent's behavior

The self-driving car analogy also suggests the human can "take the wheel at any moment," so I assumed some form of mid-run input would be available.
Why it would be nice to have
Red team work is pretty dynamic, and there are moments where being able to nudge the agent mid-run would be really useful — for example:

Adding new intel that just came up
Redirecting focus to something more interesting the stream just revealed
Asking the agent to slow down or hold before a noisy/risky action
Giving quick feedback without having to restart the run and lose context

My question
Is mid-run input something that already works and I'm just missing a hotkey or setting? Or is it planned for a future iteration? Either way, totally understandable — just wanted to flag it in case it's useful feedback.
Happy to help test or contribute if it's something the team is open to.
Thanks again!

JiwaniZakir:
Closes #54

## Summary

Enables mid-run operator input ("Human in the Loop") by allowing `submit()` in `useAgent.ts` to interrupt an active stream rather than silently dropping the message. The `Prompt` component is also unblocked so the input field remains active while the agent is streaming.

## Changes

- **`clients/cli/src/hooks/useAgent.ts` — `submit()`**: Replaced the early-return guard with an interrupt sequence: calls `clientRef.current.runs.cancelMany({ threadId, status: "running" })` to cancel backend runs, aborts the active `AbortController`, nulls `abortRef.current`, and emits a `"Interrupted by operator."` system event before proceeding with the new message.
- **`clients/cli/src/hooks/useAgent.ts` — stream cleanup**: Added an `abortController.signal.aborted` check before `resetStreamState()` so that the finishing handler of an interrupted stream does not clobber the `abortRef` or stream state belonging to the newly started run.
- **`clients/cli/src/screens/REPL.tsx`**: Changed `<Prompt isDisabled={agent.isStreaming}>` to `isDisabled={false}`, keeping the input field enabled at all times.

## Testing

- [ ] `make lint` passes
- [ ] `make test-local` passes
- [ ] Docker build succeeds (`make build`)
- [x] Manual testing (describe below)

Started a long-running agent stream, typed a follow-up message mid-run, and confirmed: (1) the `"Interrupted by operator."` system event appears in the REPL, (2) the prior stream is cancelled, and (3) the new message is submitted and a fresh run begins without dangling tool call errors. Also verified that normal (non-interrupted) runs still clean up `abortRef` and reset stream state correctly after completion.

## Related Issues

Addresses #54 — mid-run input was silently blocked; `PatchToolCallsMiddleware` on the backend handles any tool calls left open by the interruption.

---
*This PR was created with AI assistance (Claude). The changes were reviewed by quality gates and a critic model before submission.*

### DIFF
--- a/clients/cli/src/hooks/useAgent.ts
+++ b/clients/cli/src/hooks/useAgent.ts
@@ -141,9 +141,20 @@ export function useAgent({
 
   const submit = useCallback(
     (message: string): void => {
-      // Block submit while a stream is active to prevent race conditions
-      // where a second run cancels the first, leaving dangling tool calls
-      if (abortRef.current) return;
+      // If a run is already active, interrupt it so the operator can inject
+      // a message mid-run (Human in the Loop).  PatchToolCallsMiddleware on
+      // the backend handles any dangling tool calls left by the interruption.
+      if (abortRef.current) {
+        const threadId = threadIdRef.current;
+        if (threadId) {
+          clientRef.current.runs
+            .cancelMany({ threadId, status: "running" })
+            .catch(() => {});
+        }
+        abortRef.current.abort();
+        abortRef.current = null;
+        addEvent({ type: "system", content: "Interrupted by operator." });
+      }
 
       addEvent({ type: "user", content: message });
 
@@ -390,8 +401,13 @@ export function useAgent({
           setError(msg);
         }
 
-        abortRef.current = null;
-        resetStreamState();
+        // Only reset if this run was not interrupted by a new submit().
+        // An interrupt sets abortController.signal.aborted and immediately
+        // starts a fresh runStream(), so we must not clobber its state.
+        if (!abortController.signal.aborted) {
+          abortRef.current = null;
+          resetStreamState();
+        }
       };
 
       runStream().catch((err) => {
--- a/clients/cli/src/screens/REPL.tsx
+++ b/clients/cli/src/screens/REPL.tsx
@@ -243,7 +243,7 @@ export function REPL({ initialMessage }: REPLProps) {
           )}
 
           <Prompt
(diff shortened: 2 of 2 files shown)


PurpleCHOIms:
Thank you for the contribution and for identifying this issue! 🙏

This PR has been superseded by #63, which takes a different architectural approach:

**PR #62 (this PR):** Cancel the current run and start a new one → in-progress work is lost
**PR #63 (merged):** Pause at checkpoint via `cancel(action: "interrupt")` → state is preserved, operator can `/resume`

Additional features in #63:
- Claude Code-style message queuing (type during streaming, auto-submit on completion)
- Double Ctrl+C for hard cancel, single Ctrl+C for pause
- `/resume [feedback]` slash command
- Web client parity

Closing in favor of #63. Thanks again for raising #54!

PR #125, opened 2026-05-04
fix(tmux): recover from 'can't find pane: main' during parallel execution

### LINKED ISSUE #124: Bug: `can't find pane: main` error during parallel objective execution
## Bug Description

When launching parallel objectives (OBJ-008 LLM and OBJ-009 mobile RE), Decepticon throws a tmux error and fails to execute the tasks.

## Error

```
error> can't find pane: main
```

## Steps to Reproduce

1. Start Decepticon v1.0.18
2. Queue multiple independent objectives for parallel execution (e.g., OBJ-008 and OBJ-009)
3. Trigger parallel launch (`Now launching both in parallel.`)
4. Observe the `can't find pane: main` error

## Expected Behavior

Parallel objectives should launch in separate tmux panes/sessions without errors.

## Actual Behavior

Execution halts with `can't find pane: main`.

## Environment

- Decepticon version: 1.0.18
- OS: Linux (Ubuntu)
- Terminal: tmux 3.5a

## Analysis / Root Cause

The error `can't find pane: main` originates from **tmux**, not Decepticon's core logic. This occurs when the application attempts to target a tmux pane named `main` that does not exist in the current session context. Possible causes:

1. **Missing pane creation** — the session/pane initialization step may have failed silently before the parallel launch.
2. **Pane/session killed** — a prior crash or cleanup closed the pane/session.
3. **Wrong session target** — the command is targeting a session where `main` does not exist.
4. **Naming mismatch** — tmux uses pane IDs (`%0`) or indices (`.0`) by default; arbitrary names like `main` require explicit setting (e.g., `tmux select-pane -T main` or `tmux new-window -n main`). If Decepticon creates a pane/window but does not name it, subsequent lookups by name will fail.

## Suggested Fix

- Ensure pane/window creation names are explicitly set and consistent with lookup commands.
- Add fallback logic: if `main` pane lookup fails, create it or reinitialize the session.
- Consider using pane indices/IDs (e.g., `select-pane -t 0`) instead of names for more robust targeting.

## Additional Context

Relevant UI state at failure:
```
● 9 objectives planned. Now executing OBJ-008 (LLM) and OBJ-009 (mobile RE) — independent, can run in parallel.
● Now launching both in parallel.
● Delegate (exploit)
● Delegate (reverser)
error> can't find pane: main
```

VoidChecksum:
## Problem

When launching parallel objectives, Decepticon throws:

```
error> can't find pane: main
```

This blocks parallel execution and leaves the orchestrator unable to dispatch sub-agents.

## Root Cause

`TmuxSessionManager._initialized` is a class-level cache. When `initialize()` is called:

1. If the session name is in `_initialized`, it returns immediately without verifying the session still exists in tmux.
2. If the tmux server was killed externally (e.g. benchmark harness cleanup runs `tmux kill-server`), the session is gone but `_initialized` still contains the name.
3. Subsequent `_capture()` / `_send()` calls fail with `can't find pane: main` because tmux cannot resolve the target.
4. The recovery logic only caught `"no server running"` and `"session not found"`, not `"can't find pane"`.

### `decepticon/backends/docker_sandbox.py`

- **`initialize()`**: When the session is in `_initialized`, verify it still exists via `has-session` before returning. If the session is gone, remove it from the cache and proceed with creation.
- **`execute()` / `execute_async()`**: Added `"can't find pane"` to the recoverable error conditions.
- **Poll loops**: Detect `"can't find pane"` as session-death and invalidate the cache.
- **`_docker_tmux()`**: Detect `"can't find pane"` as server death and clear all session caches.

### `tests/unit/backends/test_session_log.py`

- `test_initialize_recreate_when_stale_cache`
- `test_execute_recovers_from_cant_find_pane`
- `test_execute_async_recovers_from_cant_find_pane`
- `test_docker_tmux_invalidates_cache_on_cant_find_pane`

## Verification

- Syntax checks pass for both modified files.
- New tests cover the stale-cache and recovery paths.

## Fixes

Fixes #124

### DIFF
--- a/decepticon/backends/docker_sandbox.py
+++ b/decepticon/backends/docker_sandbox.py
@@ -118,7 +118,11 @@ class TmuxSessionManager:
         if result.returncode != 0:
             error_msg = result.stderr or result.stdout
             # Detect tmux server death and invalidate session cache
-            if "no server running" in error_msg or "server exited" in error_msg:
+            if (
+                "no server running" in error_msg
+                or "server exited" in error_msg
+                or "can't find pane" in error_msg
+            ):
                 log.warning("tmux server died — invalidating all session caches")
                 with TmuxSessionManager._init_lock:
                     TmuxSessionManager._initialized.clear()
@@ -157,7 +161,18 @@ class TmuxSessionManager:
         """Create session if needed and inject PS1 marker (once per session)."""
         with TmuxSessionManager._init_lock:
             if self.session in TmuxSessionManager._initialized:
-                return
+                # Stale-cache guard: the tmux server may have been killed
+                # externally (e.g. benchmark harness cleanup) since we last
+                # created this session.
+                try:
+                    self._docker_tmux(["has-session", "-t", self.session], timeout=5)
+                    return
+                except RuntimeError:
+                    log.warning(
+                        "Session '%s' was in init cache but tmux says it is gone — recreating",
+                        self.session,
+                    )
+                    TmuxSessionManager._initialized.discard(self.session)
 
         session_exists = False
         try:
@@ -231,7 +246,11 @@ class TmuxSessionManager:
             baseline = self._capture()
         except RuntimeError as e:
             error_msg = str(e)
-            if "no server running" in error_msg or "session not found" in error_msg:
+            if (
+                "no server running" in error_msg
+                or "session not found" in error_msg
+                or "can't find pane" in error_msg
+            ):
                 log.warning("Session '%s' is dead — attempting recovery", self.session)
                 with TmuxSessionManager._init_lock:
                     TmuxSessionManager._initialized.discard(self.session)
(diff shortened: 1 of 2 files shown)


PurpleCHOIms:
Reviewed #124 root cause while working on this. This PR addresses the symptom by adding more recoverable tmux error strings, but the underlying issue is that Decepticon was targeting tmux sessions by mutable names and trusting the in-process _initialized cache. I opened a replacement fix from branch fix_tmux_error_classification_124 that tracks the actual tmux pane id returned by new-session -P -F '#{pane_id}', verifies the cached pane before reuse, and recreates stale sessions without relying on matching error text like `can't find pane`.\n\nValidation completed locally against the real decepticon-sandbox container:\n- unit tests: `51 passed` for session/backend tests\n- stale tmux session kill + recovery path passed\n- parallel session execution passed\n- LLM-facing `bash.ainvoke` / `bash_output.ainvoke` path passed\n\nRecommendation: supersede this PR with the pane-id based fix rather than merging the string-matching recovery.

PurpleCHOIms:
[#131](https://github.com/PurpleAILAB/Decepticon/pull/131)

Issue #175, opened 2026-05-07
[BUG] EngagementContextMiddleware workspace never initializes — "No engagement workspace is set" on all filesystem operations


abiboe:
### Description

After completing the full Soundwave planning interview (RoE, CONOPS, Deconfliction Plan all generated successfully), the engagement workspace is never initialized. Every filesystem operation from any agent fails with: "No engagement workspace is set. Filesystem tools are scoped to the active engagement and cannot access the shared /workspace root." The framework can plan but cannot execute — agents cannot read or write any files.

### Steps to Reproduce

1. Run `decepticon` — all 7 services start healthy
2. Complete Soundwave planning interview (target, RoE, CONOPS generated successfully)
3. Attempt any filesystem operation: save_opplan, write_file, read_file
4. Error: "No engagement workspace is set"
5. Tried workarounds:
   - Manually created /workspace/<slug>/ on host
   - Investigated PostgreSQL — only 'litellm' DB exists, no 'langgraph' DB
   - Typed `gdx insodus-tech` in terminal UI
6. Error persists across all attempts on two machines

### Expected Behavior

After Soundwave completes the planning interview, the engagement workspace should auto-initialize at /workspace/<slug>/. Agents should read/write files scoped to that engagement without manual steps.

### Actual Behavior

The langgraph database does not exist in the PostgreSQL container — only litellm DB is created on install. EngagementContextMiddleware has no database to read from, so workspace context is never injected into agent system prompts.

Errors observed:
- "No engagement workspace is set"
- HTTP 400: {"detail":"badly formed hexadecimal UUID string"}
- Model 'auth/claude-haiku-4-5' rate limit (429)

PostgreSQL check:
$ docker exec -it decepticon-postgres psql -U decepticon -d litellm -c "\l"
→ Only 'litellm' exists. No 'langgraph' database.
$ psql -U decepticon -d langgraph
→ FATAL: database "langgraph" does not exist

### Environment

- Decepticon version: 1.0.23
- OS 1: Windows 11 + WSL2 (Ubuntu) — reproduced
- OS 2: Parrot OS native — reproduced independently
- Docker: Docker Desktop latest
- Install method: curl -fsSL https://decepticon.red/install | bash
- LLM: Anthropic Claude Code OAuth (DECEPTICON_AUTH_CLAUDE_CODE=true)
- Model profile: eco

### Logs

```shell
No engagement workspace is set. Filesystem tools are scoped to the active engagement.
[...]

Issue #173, opened 2026-05-07
Filesystem tools fail with "No engagement workspace is set" after Soundwave interview despite following documentation


manch1915:
### Description

I followed the official Decepticon documentation and created a new authorized engagement through the normal Soundwave flow. The interview completed successfully, and Soundwave started generating the engagement documents.

However, when Soundwave tries to write the generated documents to the workspace, all filesystem tools fail with the following error:

No engagement workspace is set. Filesystem tools are scoped to the active engagement and cannot access the shared /workspace root.

`● Perfect. All dimensions are clear. I'll now generate the three engagement documents.

● Write (plan/roe.json)
  ⎿  No engagement workspace is set. Filesystem tools are scoped to the active engagement and cannot access the shared /workspace root.

● Let me check the workspace structure:

● List (/)
  ⎿  Error: No engagement workspace is set. Filesystem tools are scoped to the active engagement and cannot access the shared /workspace root.

● I see the issue. Let me create the plan directory structure properly:

● Write (plan/roe.json)
  ⎿  No engagement workspace is set. Filesystem tools are scoped to the active engagement and cannot access the shared /workspace root.

● I see the issue—the workspace context isn't being recognized by the filesystem tools. Let me try a different approach and write to the engagement slug directory directly:

● Write (shtemaran/plan/roe.json)
  ⎿  No engagement workspace is set. Filesystem tools are scoped to the active engagement and cannot access the shared /workspace root.

● Let me check the actual workspace structure to understand the directory layout:

● Glob (/workspace**)
  ⎿  Error: No engagement workspace is set. Filesystem tools are scoped to the active engagement and cannot access the shared /workspace root.`

### Steps to Reproduce

1. Install Decepticon using the documented installation flow.
2. Run: decepticon onboard
3. Configure the LLM provider.
4. Start Decepticon: decepticon
5. Create a new engagement.
6. Complete the Soundwave interview.
7. Approve the generated engagement details.
8. Wait for Soundwave to write the planning documents.

### Expected Behavior

After completing the Soundwave interview, the active engagement workspace should be initialized and filesystem tools should be able to write the generated planning documents

### Actual Behavior

Soundwave tries to write the documents, but every filesystem operation fails because the active engagement workspace is not recognized.

### Environment

OS: WSL2 Ubuntu on Windows
Decepticon version: 1.0.23
Install path: ~/.decepticon
Docker: Docker Compose stack
Engagement slug: shtemaran

### Logs

```shell
[...]

Issue #152, opened 2026-05-05
OMP filesystem tools blocked: "No engagement workspace is set" when writing planning docs


NetVar1337:
## Summary (updated with root cause + fix + additional bug discovered)

When Soundwave completes operator interviews and attempts to write engagement planning documents (RoE, CONOPS, Deconfliction Plan) via OMP filesystem tools (`write`, `read`, `find`), the tools fail with:

```
No engagement workspace is set. Filesystem tools are scoped to the active engagement and cannot access the shared /workspace root.
```

However, the engagement workspace IS correctly mounted inside the sandbox at `/workspace/` (confirmed via `docker exec`).

## Root Cause

**Triple architecture mismatch:**

### Layer 1: LangGraph FilesystemMiddleware (`/app/decepticon/middleware/filesystem.py`)

The `EngagementFilesystemBackend` wraps deepagents filesystem tools and maps virtual `/workspace/` paths to the real engagement directory. Two bugs:

1. **`_normalize_engagement_workspace()`** returns `None` when `workspace_path == "/workspace"` — but the sandbox mounts the engagement directly at `/workspace` (not as a subdirectory like `/workspace/banco-plata`).

2. **`_workspace_from_runtime()`** returns `None` when `workspace_path` is not in the agent's runtime state or configurable. Sub-agents (Soundwave, recon, exploit, etc.) don't inherit the workspace_path from the parent orchestrator.

3. **💥 Additional bug discovered:** The original `_normalize_engagement_workspace` calls `DockerSandbox._normalize_workspace_path(workspace_path)` — but this **method does not exist** anywhere in the DockerSandbox class or its MRO. The call was always dead code because the function always returned `None` before reaching it (due to bug #1 above). When my initial patch actually tried to call the method, it threw `AttributeError`.

### Layer 2: OMP (Oh My Pi) tools on the HOST

The OMP agent (`omp` binary) runs directly on the **host machine**, not inside any container. Its filesystem tools (`read`, `write`, `edit`, `find`, `search`, `ast_grep`, `ast_edit`, `glob`) operate on the host filesystem, not inside the sandbox container.

- The sandbox has `/workspace` → `/home/void/.decepticon/workspace/banco-plata` mounted
- The HOST has **no** `/workspace/` directory
- OMP tools look for `/workspace/` on the host → fail

### Layer 3: `.env` configuration

[...]

PR #180, opened 2026-05-08
fix(cli): preserve soundwave-authored slug across handoff to decepticon

### LINKED ISSUE #159: Soundwave→Decepticon handoff drops engagement slug — workspace_path lost on thread transition
## Summary
When Soundwave finishes engagement planning and emits `engagement_ready`, the CLI drops the current thread and starts a new one for Decepticon. But the engagement slug captured during the handoff is **only used for logging** — it is never passed to the new thread's run input.

In `/app/src/hooks/useAgent.ts` (`handleStreamComplete`, line 526-534):

```typescript
if (pendingHandoffRef.current) {
  threadIdRef.current = null;
  lastCountRef.current = 0;
  askedQuestionIds.current.clear();
  pendingHandoffRef.current = null;  // slug discarded here
}
```

The next `submit()` call creates a fresh thread with:

```typescript
const slug = process.env.DECEPTICON_ENGAGEMENT; // NEW session, slug from env
const input = { messages: [...] };
if (slug) {
  input.engagement_name = slug;
  input.workspace_path = process.env.DECEPTICON_WORKSPACE_PATH ?? "/workspace";
}
```

Since `DECEPTICON_ENGAGEMENT` is rarely set, the new thread starts with **no engagement context** — no `engagement_name`, no `workspace_path` in the state.

## Impact
- The orchestrator (Decepticon) starts without knowing the engagement name or workspace
- `EngagementContextMiddleware` injects a blank system prompt addendum
- `EngagementFilesystemBackend` must rely on the env var fallback (`DECEPTICON_WORKSPACE_PATH`) to determine the workspace root
- If the env var is also unset, all filesystem tools fail with "No engagement workspace is set"

## Fix
Capture the slug during handoff and pass it to the next submit:

```typescript
// Store slug for next submit
const queuedSlugRef = useRef<string | null>(null);

// In handleStreamComplete:
if (pendingHandoffRef.current) {
  queuedSlugRef.current = pendingHandoffRef.current;
  threadIdRef.current = null;
  ...
}

// In submit():
const slug = queuedSlugRef.current || process.env.DECEPTICON_ENGAGEMENT;
input.engagement_name = slug;
input.workspace_path = process.env.DECEPTICON_WORKSPACE_PATH ?? "/workspace";
```

## Steps to Reproduce
1. Start a new engagement via Soundwave
2. Let Soundwave complete the planning interview and emit `engagement_ready`
3. Wait for the handoff to Decepticon
4. Check the new thread's state — `engagement_name` and `workspace_path` are absent

## Related
- #156 — Missing DECEPTICON_ENGAGEMENT env var

Lempkey:
## Summary

- Adds `queuedEngagementSlugRef` to capture the slug emitted by Soundwave's `complete_engagement_planning` tool via the `engagement_ready` custom event (`useAgent.ts:303`)
- Extracts `buildSubmitInput()` as a pure helper that applies slug precedence: the queued handoff slug wins over `DECEPTICON_ENGAGEMENT` env var, which may be stale or empty for engagements created during the Soundwave interview
- Clears the queued slug in `handleStreamComplete` (alongside the existing `pendingHandoffRef` clear) so it is not re-injected on subsequent Decepticon submits on the same thread
- Leaves `pendingHandoffRef` boolean semantics untouched; the handoff still fires even when the `engagement_ready` event carries no slug

## Root cause

`pendingHandoffRef` captured the slug at L303 but the slug string was discarded at L535 (`pendingHandoffRef.current = null`) before the next Decepticon submit at L687 could use it. The submit fell back to `process.env.DECEPTICON_ENGAGEMENT`, which is the launcher-set value and is stale when the operator created a new engagement during the interview.

## Test plan

- [x] `npm run typecheck` — TypeScript clean
- [x] `npm test` — 8 new tests in `buildSubmitInput.test.ts` pass, 24 total pass
- [ ] End-to-end: run a full Soundwave interview (`make dev` + `make cli`), complete planning, verify the first Decepticon message includes the correct `engagement_name` (requires full stack — impractical for CI)

## Note on downstream failures

Filesystem tool failures (#152, #173, #175) are downstream symptoms that still require a **follow-up PR** fixing `_normalize_engagement_workspace` at `decepticon/middleware/filesystem.py:32` — that function currently rejects `/workspace` as "no engagement workspace" and will continue to do so until repaired. This PR fixes the slug-drop at the handoff point; the middleware fix unblocks the filesystem tools.

Closes #159

### DIFF
--- a/clients/cli/src/hooks/buildSubmitInput.test.ts
+++ b/clients/cli/src/hooks/buildSubmitInput.test.ts
@@ -1,0 +1,102 @@
+import { describe, expect, it } from "vitest";
+import { buildSubmitInput } from "./buildSubmitInput.js";
+
+describe("buildSubmitInput", () => {
+  it("always includes the user message", () => {
+    const { input } = buildSubmitInput({
+      message: "hello",
+      handoffSlug: null,
+      envSlug: undefined,
+      envWorkspacePath: undefined,
+      modelOverride: null,
+    });
+    expect(input.messages).toEqual([{ role: "user", content: "hello" }]);
+  });
+
+  describe("slug precedence", () => {
+    it("uses handoffSlug when present, ignoring envSlug — regression #159", () => {
+      const { input } = buildSubmitInput({
+        message: "go",
+        handoffSlug: "soundwave-authored",
+        envSlug: "launcher-set",
+        envWorkspacePath: "/workspace",
+        modelOverride: null,
+      });
+      expect(input.engagement_name).toBe("soundwave-authored");
+    });
+
+    it("falls back to envSlug when handoffSlug is null", () => {
+      const { input } = buildSubmitInput({
+        message: "go",
+        handoffSlug: null,
+        envSlug: "from-env",
+        envWorkspacePath: undefined,
+        modelOverride: null,
+      });
+      expect(input.engagement_name).toBe("from-env");
+      expect(input.workspace_path).toBe("/workspace");
+    });
+
+    it("emits no engagement fields when both are null/undefined", () => {
+      const { input } = buildSubmitInput({
+        message: "go",
+        handoffSlug: null,
+        envSlug: undefined,
+        envWorkspacePath: undefined,
+        modelOverride: null,
+      });
+      expect(input.engagement_name).toBeUndefined();
+      expect(input.workspace_path).toBeUndefined();
+    });
+  });
+
+  describe("workspace_path", () => {
+    it("uses envWorkspacePath when slug wins", () => {
+      const { input } = buildSubmitInput({
+        message: "go",
+        handoffSlug: "eng",
+        envSlug: undefined,
+        envWorkspacePath: "/custom/path",
+        modelOverride: null,
+      });
+      expect(input.workspace_path).toBe("/custom/path");
+    });
+
+    it("defaults workspace_path to /workspace when envWorkspacePath is undefined", () => {
+      const { input } = buildSubmitInput({
+        message: "go",
+        handoffSlug: null,
+        envSlug: "from-env",
+        envWorkspacePath: undefined,
+        modelOverride: null,
+      });
+      expect(input.workspace_path).toBe("/workspace");
+    });
+  });
+
+  describe("modelOverride", () => {
+    it("injects modelOverride into input and streamConfig.configurable", () => {
+      const { input, streamConfig } = buildSubmitInput({
+        message: "go",
+        handoffSlug: null,
+        envSlug: undefined,
+        envWorkspacePath: undefined,
+        modelOverride: "claude-opus-4",
+      });
+      expect(input.model_override).toBe("claude-opus-4");
+      expect(streamConfig.configurable?.model_override).toBe("claude-opus-4");
+    });
+
+    it("emits no model_override fields when null", () => {
+      const { input, streamConfig } = buildSubmitInput({
+        message: "go",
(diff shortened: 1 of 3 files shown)


Lempkey:
Thanks for the thorough review. The high-severity finding was correct — \`queuedEngagementSlugRef\` was being nulled inside \`handleStreamComplete()\` (during the Soundwave teardown), which ran *before* the auto-submitted Decepticon \`submit()\` could consume it. The ref arrived as \`null\` at \`buildSubmitInput\`, so it always fell back to the stale env var.

**Fix (7247f00):**

Two changes in \`useAgent.ts\`:

1. **Removed** \`queuedEngagementSlugRef.current = null\` from \`handleStreamComplete\`. A comment explains the intent: the ref must survive until the Decepticon run consumes it.

2. **In \`submit()\`**, capture the slug into a local *before* the stream starts, then clear the ref only after \`processStream()\` resolves successfully:

\`\`\`typescript
const handoffSlug = queuedEngagementSlugRef.current;
const { input, streamConfig } = buildSubmitInput({ handoffSlug, ... });

try {
  const stream = client.runs.stream(...);
  await processStream(stream, abortController);
  if (handoffSlug) queuedEngagementSlugRef.current = null; // clear after success
} catch (err) {
  // slug retained on failure — safe to retry
  ...
}
\`\`\`

**On the lifecycle test gap:** Agreed the existing tests don't cover the \`engagement_ready → handleStreamComplete → next submit\` sequence. Adding those requires \`@testing-library/react\` + jsdom (the current vitest config uses \`environment: "node"\`), which is out of scope for this fix. Happy to follow up in a separate PR if the maintainers want to expand the test infrastructure.

PurpleCHOIms:
After re-examining the root cause of this regression (#159), I concluded that the architecture itself — **letting the LLM be the source of truth for the engagement slug** — is the actual problem. Replacing this PR with #182 (`refactor/engagement-handoff-signal-only`), which takes the opposite approach.

## Why the redirect

The natural product flow:
1. **Launcher** (Go picker UI / web env) decides the engagement slug — the user picks or types it once at session start.
2. **Soundwave** writes RoE/CONOPS/Deconfliction into the already-mounted `/workspace/plan/`.
3. **Decepticon** picks up the same `/workspace` after handoff.

Under this flow the slug is launcher-owned. The slug Soundwave's LLM emits via `complete_engagement_planning(engagement_name=…)` is just an echo, and there is no reason to treat it as authoritative. This PR's `queuedEngagementSlugRef` + queue/clear lifecycle was designed to privilege the LLM-emitted slug, but if the launcher-mounted directory and the LLM-emitted slug ever diverge (hallucination, normalization variance), state and filesystem can mismatch — a worse failure mode than the original drop.

## What #182 does instead

- Drops the `engagement_name` argument from `complete_engagement_planning` → pure boolean handoff signal.
- The CLI injects launcher's env-set slug into `config.configurable` on every submit.
- `EngagementContextMiddleware.before_agent` hydrates configurable into state.
- The "slug drop" bug class no longer exists — there is no slug channel to drop from.
- `queuedEngagementSlugRef` becomes unnecessary.

## What survives

Your regression tests (`buildSubmitInput.test.ts`) and root-cause analysis were spot on — #159 was diagnosed correctly here, and #182 was designed so that the same issue closes. If you'd like to dispatch follow-up work after this closes, that would be welcome.

Closing this PR.

PR #181, opened 2026-05-08
fix(middleware): resolve /workspace correctly under launcher mode

Lempkey:
## Summary

- `_normalize_engagement_workspace()` was collapsing `workspace_path="/workspace"` to `None`, causing every filesystem tool (`ls`, `read`, `write`, `edit`, `grep`, `glob`, `download_files`) to return `"No engagement workspace is set..."` even when an engagement is active.
- The bug: the function conflated "path resolved to the literal `/workspace`" with "no engagement configured", which is only true in raw `make dev` mode. In launcher mode the Docker bind-mount already scopes `/workspace` to the engagement, so `/workspace` IS the correct engagement root.
- Fix: rewrite the normalizer to accept `/workspace` explicitly, accept `/workspace/<safe-slug>` (unchanged), and fail closed for `None`/empty/traversal/invalid paths. Invalid paths are not silently coerced into `/workspace`.

Prerequisite: #180 (slug propagation) must be merged first — this PR fixes the second half of the chain.

## What changed

**`decepticon/middleware/filesystem.py`**: Rewrote `_normalize_engagement_workspace` with a strict 4-case contract: empty→`None`, `/workspace`→`/workspace`, `/workspace/<slug>`→normalized (only if `_normalize_workspace_path` did not silently coerce), anything else (including traversal like `/workspace/../etc`)→`None`. Added `os.path.normpath` guard to catch `..`-based traversal that the slug regex would otherwise accept. Added `import os`.

**`tests/unit/middleware/test_filesystem.py`**: Flipped `test_root_workspace_fails_closed` → `test_root_workspace_accepted_as_engagement_root`. Added: no-prefix-doubling under `/workspace` root, glob scoping, trailing-slash acceptance, explicit negative regression that invalid paths fail closed and do NOT coerce to `/workspace`.

## Test plan

- [x] `pytest tests/unit/middleware/test_filesystem.py -v` — 11/11 pass
- [x] `make test-local` — 726 passed, 1 skipped (pre-existing skip), 0 failures
- [x] `ruff check` + `ruff format --check` on changed files — clean
- [x] `basedpyright decepticon/middleware/filesystem.py` — 0 errors
- [ ] Manual smoke (launcher mode `ls /workspace` from an active engagement) requires the LangGraph + Docker stack and is out of scope for CI

## Notes

[...]

PR #182, opened 2026-05-09
refactor(engagement): launcher-authoritative slug via config.configurable

### LINKED ISSUE #152: OMP filesystem tools blocked: "No engagement workspace is set" when writing planning docs

PurpleCHOIms:
## Summary

Move `engagement_name` and `workspace_path` from input state to runnable `config.configurable`. The **launcher** (and web `terminal-server` env) is the single source of truth; the LLM no longer decides the engagement slug.

This supersedes both **#180** (slug queuing in CLI) and **#181** (filesystem `_normalize_engagement_workspace` strict contract). #181's contract is incorporated wholesale; #180's bug class no longer exists once the slug stops traveling through the LLM.

## Why this direction

The natural product flow is:
1. Launcher decides the engagement slug (picker UI / web env)
2. Soundwave runs the interview and writes `/workspace/plan/{roe,conops,deconfliction}.json`
3. CLI flips `assistant_id` to `decepticon` on `engagement_ready`
4. Decepticon reads the planning bundle from the shared `/workspace` mount

In the previous architecture the LLM was the source of truth for the slug — `complete_engagement_planning(engagement_name=…)` echoed a slug back via a custom event, the CLI captured it, then dropped it before the next submit (#159). The fix could either rescue the LLM-emitted slug (#180) or remove the LLM from the slug path entirely. This PR takes the second option, which is the LangGraph-native shape: launcher → `config.configurable` → middleware hydration → state.

## Backend changes

- **`decepticon/tools/interaction/complete_planning.py`** — drop the `engagement_name` argument. The tool is now a pure boolean handoff signal; the emitted `engagement_ready` event carries no slug.
- **`decepticon/middleware/engagement.py`** — add `before_agent` hydration that copies `engagement_name` and `workspace_path` from `config.configurable` into state on the first agent step. Idempotent: state values win when present. Downstream middleware (OPPLAN, filesystem) read state as before, so no other middleware needs to change.
- **`decepticon/middleware/filesystem.py`** — `_normalize_engagement_workspace` rewritten with a strict 4-case contract (empty → `None`, `/workspace` accepted as the engagement root for launcher mode, `/workspace/<safe-slug>` normalized, anything else including `..` traversal → `None` with no silent coercion). Adapted from #181's diagnosis.

## Frontend changes

- **`clients/cli/src/hooks/useAgent.ts`** — send `engagement_name` / `workspace_path` via `config.configurable` on every run instead of `input` state fields. `pendingHandoffRef` downgraded from `string | null` to `boolean` (no slug to retain). `model_override` is now passed through `configurable` only (input field removed; ModelOverrideMiddleware reads either path, but a single channel is cleaner).
- **`clients/shared/streaming/src/types.ts`** — drop the `engagement?` field from `SubagentCustomEvent`.
- **`decepticon/agents/prompts/soundwave.md`** — the Completion Signal section no longer instructs the LLM to pass a slug to the tool.

## Tests

- `tests/unit/middleware/test_engagement.py` — +6 tests covering `_hydrate_engagement_state` (empty state pulls from configurable, idempotent when state already set, partial hydration, neither present, defensive non-string filtering) and `_resolve_workspace_path` precedence.
- `tests/unit/middleware/test_filesystem.py` — +5 tests for the new contract: `/workspace` accepted as engagement root, no prefix doubling under root, trailing-slash variant accepted, traversal fails closed, paths outside `/workspace` fail closed. The previous `test_root_workspace_fails_closed` is flipped to `test_root_workspace_accepted_as_engagement_root` to match the new launcher-mode semantics.

## Verification

```bash
uv run ruff check .                                  # clean
uv run ruff format --check decepticon/ tests/        # clean
uv run basedpyright --level error                    # 0 errors
uv run pytest -n auto -q -m \"not slow\"             # 724 passed, 0 failed
cd clients/cli && npm run typecheck && npm test      # tsc clean, 16/16 pass
cd clients/cli && npm run build                      # tsc clean
```

## Closes / Supersedes

Closes #152, #159, #173. Addresses the filesystem failure in #175.

Supersedes PR #180 (slug queuing — bug class no longer exists). Supersedes PR #181 (filesystem normalize — contract incorporated). Note: PR #178's filesystem fix-1 is rejected on factual grounds — the diagnosis that \`DockerSandbox._normalize_workspace_path\` does not exist is incorrect; the static method is at \`docker_sandbox.py:830\` and is used in 11 places. PR #178's subagent_streaming fix-2 (None tool_call_id guard) is independent and should land separately.

## Test plan

- [x] Python lint + typecheck + full unit test suite locally
- [x] CLI typecheck + unit tests + build locally
- [ ] End-to-end: \`make dogfood\` — Soundwave interview → handoff → Decepticon fs tool (\`ls /workspace\`) — requires the full stack and is appropriate for the dogfood gate before tagging

### DIFF
--- a/clients/cli/src/hooks/useAgent.ts
+++ b/clients/cli/src/hooks/useAgent.ts
@@ -152,9 +152,11 @@ export function useAgent({
   // tool flips this to "decepticon" mid-flight; the next submit() then opens
   // a fresh thread on the new assistant.
   const assistantIdRef = useRef<string>(INITIAL_ASSISTANT_ID);
-  // Slug captured from the engagement_ready event — kept for system-level
-  // logging when the handoff fires.
-  const pendingHandoffRef = useRef<string | null>(null);
+  // Boolean handoff signal — set when soundwave emits engagement_ready; consumed
+  // in handleStreamComplete to drop the soundwave thread before the auto-submit
+  // opens a fresh decepticon thread. Carries no slug; the launcher is the single
+  // source of truth and reaches the agent via config.configurable.
+  const pendingHandoffRef = useRef<boolean>(false);
 
   // Derived for backward compatibility
   const isStreaming = runState === "streaming" || runState === "connecting";
@@ -299,15 +301,15 @@ export function useAgent({
             // active assistant so the next submit() lands on decepticon.
             // The current run continues to completion (soundwave's closing
             // message); thread handoff fires from handleStreamComplete.
-            const slug = data.engagement ?? "";
-            pendingHandoffRef.current = slug || "(unnamed)";
+            // Pure boolean signal — the engagement slug travels independently
+            // via config.configurable from the launcher's env.
+            pendingHandoffRef.current = true;
             assistantIdRef.current = "decepticon";
             setAssistantId("decepticon");
             addEvent({
               type: "system",
-              content: slug
-                ? `Engagement '${slug}' planning complete — Decepticon will pick up your next message.`
-                : "Engagement planning complete — Decepticon will pick up your next message.",
+              content:
+                "Engagement planning complete — Decepticon will pick up your next message.",
             });
             break;
           }
@@ -532,7 +534,7 @@ export function useAgent({
         threadIdRef.current = null;
         lastCountRef.current = 0;
         askedQuestionIds.current.clear();
-        pendingHandoffRef.current = null;
+        pendingHandoffRef.current = false;
       }
 
       // Auto-submit queued message
@@ -679,31 +681,31 @@ export function useAgent({
         setActiveAgent("decepticon");
         setStreamStats({ startTime: Date.now(), totalTokens: 0, promptTokens: 0, completionTokens: 0 });
 
-        // Engagement context flows in as regular state fields. A small
-        // middleware on the agent side picks engagement_name + workspace_path
-        // out of state and injects them into the model's system prompt —
-        // the LangGraph-idiomatic way to surface launcher-set context to the
-        // LLM without polluting user messages.
-        const slug = process.env.DECEPTICON_ENGAGEMENT;
+        // Engagement context and the /model override flow as runnable
+        // ``config.configurable`` entries. EngagementContextMiddleware
+        // hydrates state from configurable on before_agent so OPPLAN and
+        // filesystem middlewares see the values as ordinary state fields,
+        // and ModelOverrideMiddleware reads model_override straight from
+        // configurable. The launcher is the single source of truth for the
+        // engagement slug; the LLM never decides it.
         const input: Record<string, unknown> = {
           messages: [{ role: "user", content: message }],
         };
+
+        const configurable: Record<string, unknown> = {};
+        const slug = process.env.DECEPTICON_ENGAGEMENT;
         if (slug) {
-          input.engagement_name = slug;
-          input.workspace_path = process.env.DECEPTICON_WORKSPACE_PATH ?? "/workspace";
+          configurable.engagement_name = slug;
+          configurable.workspace_path =
+            process.env.DECEPTICON_WORKSPACE_PATH ?? "/workspace";
+        }
+        const modelOverride = getModelOverride();
+        if (modelOverride) {
+          configurable.model_override = modelOverride;
         }
 
-        // /model command — inject the active runtime override into BOTH
-        // input state and config.configurable. The agent's
-        // ModelOverrideMiddleware reads either path, so the call works
-        // whether the LangGraph runtime surfaces the field via state or
-        // via the runnable config.
-        const modelOverride = getModelOverride();
-        const streamConfig: { configurable?: Record<string, unknown> } = {};
-        if (modelOverride) {
-          input.model_override = modelOverride;
-          streamConfig.configurable = { model_override: modelOverride };
-        }
+        const hasConfigurable = Object.keys(configurable).length > 0;
+        const streamConfig = hasConfigurable ? { configurable } : undefined;
 
         try {
           const stream = client.runs.stream(
@@ -711,7 +713,7 @@ export function useAgent({
             assistantIdRef.current,
             {
               input,
-              ...(modelOverride ? { config: streamConfig } : {}),
+              ...(streamConfig ? { config: streamConfig } : {}),
               ...STREAM_OPTIONS,
               onDisconnect: "continue",
               signal: abortController.signal,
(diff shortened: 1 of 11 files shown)


PurpleCHOIms:
PR scope updated:

1. **CI fix** — the first push failed `ruff format --check .` because of pre-existing format drift in `scripts/render_benchmark_charts.py` on main, unrelated to the refactor. Cleaned up in a separate commit (`style(scripts): ruff format render_benchmark_charts.py`). Whitespace-only.
2. **PR #178 fix-2 absorbed** — the None `tool_call_id` guard in `subagent_streaming.py` was incorporated into this PR (commit `fix(subagent_streaming): guard None tool_call_id from corrupting active dict`). Changes:
   - Guard `active_tool_calls[tc["id"]] = tc`: skip when `tc.get("id") is None`, plus an observability warning log.
   - Tighter type: `dict[str, dict]` → `dict[str, ToolCall]` (the actual TypedDict).
   - Three regression tests: None-id storage skipped, ToolMessage falls through to "unknown" on miss, and a valid id is unaffected by a concurrent None-id call.

   The diagnosis in PR #178 ("results dropped") was inaccurate — the practical effect is wrong-name attribution rather than a drop. The guard prevents that corruption and converts it into graceful degradation.

Verification:
- `uv run pytest -n auto -q -m "not slow"`: **727 passed** (3 new)
- `uv run ruff check .`, `ruff format --check .`, `basedpyright --level error`: all clean
- `npm run typecheck`, `npm test`, `npm run build` (CLI): all pass

Lempkey closed PR #178, but the value of fix-2 survives here. Thanks for the regression analysis.

PR #193, opened 2026-05-11
test(updater): coverage for SelfUpdate, WriteVersion, downloadFile, ApplyUpdate

Lempkey:
## Summary

Expands the updater test suite introduced alongside #188 (interactive update prompt + re-exec). That PR added \`PromptIfUpdateAvailable\` and \`ApplyUpdate\` but left the functions they delegate to — \`SelfUpdate\`, \`SyncConfigFiles\`, \`downloadFile\`, \`WriteVersion\` — with zero unit coverage. This PR pins those paths so a future refactor cannot silently regress them.

This is a test-only PR. No production behaviour is changed.

## Changes

- **\`updater.go\`**: extract \`os.Executable\` into a package-level \`executableFn\` var so tests can redirect binary writes to a temp dir instead of clobbering the running test binary. Matches the \`isWSLFn\`/\`wslHostIPFn\` injection pattern already established in \`cmd/start.go\`.

- **\`updater_test.go\`**: 10 new tests (15 total, up from 5):

  | Group | Tests |
  |-------|-------|
  | \`downloadFile\` | success (content verified), HTTP error (500 → error), network error (closed server), parent-directory auto-creation |
  | \`SelfUpdate\` | no matching asset for current platform (GOOS/GOARCH named in error), HTTP 403 on download, successful write + atomic rename via \`executableFn\` override (asserts \`.tmp\` cleaned up) |
  | \`WriteVersion\` | strips \`v\`-prefix, preserves bare version string |
  | \`ApplyUpdate\` | \`SelfUpdate\` failure propagated as \`"binary update: …"\` error while \`SyncConfigFiles\`/\`compose.Pull\` failures remain non-fatal warnings |

## Testing

- [x] \`go test ./internal/updater/... -v\` — 15 tests pass, 0 failures
- [x] \`go test ./...\` (full launcher suite) — all packages green
- [ ] \`make smoke\` — not applicable (test-only PR, no runtime behaviour change)

## Related

Follow-up to #188 per the pattern of test coverage for new launcher features.

### DIFF
--- a/clients/launcher/internal/updater/updater.go
+++ b/clients/launcher/internal/updater/updater.go
@@ -159,7 +159,7 @@ func SelfUpdate(release *Release) error {
 	}
 
 	// Write to temp file first
-	execPath, err := os.Executable()
+	execPath, err := executableFn()
 	if err != nil {
 		return fmt.Errorf("get executable path: %w", err)
 	}
@@ -347,6 +347,11 @@ func PromptIfUpdateAvailable(currentVersion string) (bool, error) {
 	// child, so this is also unreachable on Windows. Kept for symmetry.
 	return true, nil
 }
+
+// executableFn is a var so tests can redirect binary writes to a temp dir
+// instead of overwriting the test binary itself. Matches the isWSLFn /
+// wslHostIPFn pattern used in cmd/start.go.
+var executableFn = os.Executable
 
 // isInteractiveStdin returns true when the launcher's stdin is connected
 // to a real terminal. Piped / redirected stdin (CI, log shippers,
--- a/clients/launcher/internal/updater/updater_test.go
+++ b/clients/launcher/internal/updater/updater_test.go
@@ -2,8 +2,13 @@ package updater
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


PurpleCHOIms (approved):
LGTM. The `executableFn` injection pattern matches the existing `isWSLFn`/`wslHostIPFn` convention in `cmd/start.go`, and the 10 new tests pin the three behavioral contracts that matter most:

- `WriteVersion` v-prefix stripping (downstream `.version` consumers depend on this format)
- `SelfUpdate` error message naming the current GOOS/GOARCH (operator diagnostic)
- `ApplyUpdate` SelfUpdate-only-fatal semantics (matches the function's documented contract at lines 226-230)

The `t.Cleanup` + `t.TempDir` + `httptest.NewServer` patterns are idiomatic, and CI is green across all targets. Confirmed that the tested functions are shared between the explicit `decepticon update` command and the interactive launch-time prompt, so the coverage remains load-bearing regardless of which entry path the operator uses.

Non-blocking nits worth considering for a follow-up:
- Move `executableFn` to the top of `updater.go` near the other package-level declarations (`Repo`/`APIBaseURL`/`RawBaseURL`) — currently sitting between two functions.
- A test for mid-copy failure cleanup (the `tmp.Close(); os.Remove(tmpPath)` path in `SelfUpdate`).
- `TestApplyUpdate_SelfUpdateErrorPropagates` exercises the propagation through a side-effect (asset platform mismatch). A direct mock of `SelfUpdate` would pin the contract more precisely, but the current approach is fine.

Thanks for the thorough follow-up to #188.

PR #194, opened 2026-05-11
fix(launcher): replace cp -r with Go-native recursive copy in backupWorkspace

Lempkey:
## Summary

Fixes a silent data-loss bug in the \`decepticon remove\` uninstall flow on Windows.

When \`backupWorkspace\` cannot use \`os.Rename\` (cross-device move), it fell back to:

\`\`\`go
exec.Command("cp", "-r", src, dst)
\`\`\`

\`cp\` does not exist on Windows. The command fails silently, the workspace is never backed up, and \`os.RemoveAll\` then deletes it — the user loses their engagement data with no warning.

## Changes

- **\`cmd/remove.go\`**: replace the \`exec.Command("cp", ...)\ ` fallback with a pure-Go recursive copy (\`copyDirAll\` / \`copyFile\`) using \`filepath.Walk\` + \`io.Copy\`. File permissions are preserved. No new dependencies — stdlib only. The \`os.Rename\` fast path (same-device) is unchanged.

- **\`cmd/remove_test.go\`** (new file, 12 tests — zero existed before):

  | Group | Tests |
  |-------|-------|
  | \`backupWorkspace\` | rename succeeds (src gone, dst has files), target already exists (error) |
  | \`copyDirAll\` | single file, nested directories, empty subdirectory, file permission preservation |
  | \`cleanPathFromFile\` | removes bash block, removes fish block, strips preceding blank line, preserves unrelated PATH lines, no-op on missing file, no-op when no changes needed |

## Testing

- [x] \`go test ./cmd/... -v\` — 12 new tests pass, 7 existing tests pass (19 total)
- [x] \`go test ./...\` (full launcher suite) — all packages green
- [ ] \`make smoke\` — not applicable; backup only runs during \`decepticon remove\` which requires the full Docker stack

## Notes

The \`cleanPathFromFile\` tests are included as a bonus — that function had zero coverage and lives in the same file. The fix is scoped to the cross-platform bug; Windows shell rc cleanup (\`cleanShellRC\` has no PowerShell support) is a separate concern left for a follow-up.

### DIFF
--- a/clients/launcher/cmd/remove.go
+++ b/clients/launcher/cmd/remove.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"bufio"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


Lempkey:
Closing this PR — the docs explicitly state that native Windows is not supported and that Windows users must run Decepticon inside WSL2. Since WSL2 is Linux, `cp` is always available and the described data-loss scenario cannot be triggered on any supported platform. The bug report was based on a misread of the supported OS matrix. Thanks for the review feedback!

PR #203, opened 2026-05-12
fix(launcher/soundwave): auto-stage source_code targets into sandbox workspace

### LINKED ISSUE #202: bug(soundwave/launcher): local-path targets not staged into sandbox — analyst falls back to black-box with no warning
## Summary

When a local filesystem path is given as the target during the Soundwave interview, the path is recorded in the RoE but **never copied into the sandbox container**. The analyst agent receives the path via the RoE but cannot reach it — the host filesystem is not visible inside the container — so it silently falls back to black-box-only analysis with no error or warning to the operator.

\`docs/engagement-workflow.md\` explicitly lists **Local path** (\`/path/to/target/app\`) as a supported target type, so this is a documented capability gap, not a missing feature.

---

## Steps to Reproduce

1. Start a new engagement via \`make dogfood\` (or \`decepticon\`).
2. During the Soundwave interview, enter a local filesystem path as the target (e.g. \`/Users/<you>/Projects/myapp/services/api\`).
3. Approve the generated RoE / ConOps / OPPLAN.
4. Let the engagement run. The analyst agent starts OBJ-002 (source code static analysis).

**Expected:** Analyst finds source at a well-known path inside the sandbox (e.g. \`/workspace/src\`) and performs white-box analysis.

**Actual:** Analyst logs something like:

\`\`\`
The host path is not mounted in this container.
Confirming this is a genuine engagement issue: the RoE specifies the source
path as /Users/... but the sandbox can't reach it.
\`\`\`

OBJ-002 is blocked. OBJ-003 (exploit) proceeds with black-box only — severely degrading the engagement quality.

---

## Root Cause

Three layers of the stack each know about the local path but none of them act on it:

### 1. Soundwave — planning-only, no file ops (by design)

\`soundwave.py\` and its skills (\`ask_user.py\`, \`complete_planning.py\`, etc.) write planning documents (RoE/ConOps/OPPLAN) but have no file-copy or workspace-initialization logic. Rule 1 of the Soundwave skill: "No Execution." Correct by design — but the hand-off to a staging step never happens.

### 2. Launcher — no source staging step

\`clients/launcher/internal/engagement/picker.go\` creates the engagement workspace and sets \`DECEPTICON_ENGAGEMENT_WORKSPACE\`, but after Soundwave finishes and before the autonomous loop starts, there is no step that reads the RoE for \`local_path\` / \`source_code\` target types and copies the source tree into the workspace.

### 3. \`docker-compose.yml\` — static workspace mount only

\`\`\`yaml
# sandbox service
volumes:
  - \${DECEPTICON_ENGAGEMENT_WORKSPACE:-\${DECEPTICON_HOME:-~/.decepticon}/workspace}:/workspace
\`\`\`

The workspace is bound at compose-up time. There is no dynamic bind-mount injection per engagement, and no path for a local source tree to reach \`/workspace/src\` automatically.

### Existing primitive that could be reused

\`decepticon/backends/docker_sandbox.py\` already implements \`upload_files()\` using \`docker cp\`:

\`\`\`python
def upload_files(self, files: list[tuple[str, bytes]]) -> list[FileUploadResponse]:
    # Writes content via docker cp into the sandbox
    subprocess.run(["docker", "cp", tmp_path, f"{self._container_name}:{path}"])
\`\`\`

This is the right tool — it just needs to be called during engagement initialization for local-path targets.

---

## Impact

- **White-box engagements silently degrade to black-box.** The operator gets no error; they only discover the problem when the analyst reports the blocker after spending time on recon.
- **Auth schemes are opaque without source.** In a real case: the analyst ran 107+ black-box probes against a FastAPI service and couldn't find the auth header because it used a non-standard name (\`X-Internal-Key\`) that was only visible in the client-side source code. White-box access would have found this immediately.
- **Affects the VPS install path too.** Even if the operator's source tree lives on a remote host, there's no way to stage it short of manual \`docker cp\`.

---

## Proposed Fix

A pre-execution staging hook in the launcher, after Soundwave finishes and before the autonomous loop starts:

**Pseudocode (launcher, Go):**
\`\`\`go
// After engagement documents are approved:
roe := loadROE(engagementWorkspace)
for _, scope := range roe.Scope {
    if scope.Type == "local_path" || scope.Type == "source_code" {
        srcDir := scope.Target   // e.g. /Users/.../myapp
        dstDir := filepath.Join(engagementWorkspace, "src")
        if err := stagePath(srcDir, dstDir); err != nil {
            log.Warn("Could not stage source path — continuing without white-box access", "path", srcDir, "err", err)
        }
    }
}
\`\`\`

Because \`/workspace\` is already bind-mounted to the engagement workspace, anything copied to \`<engagementWorkspace>/src/\` is immediately visible inside the sandbox at \`/workspace/src/\`.

**Alternative (decepticon skill):** A \`setup_source\` skill that runs as the first step of OBJ-002, reads the RoE's local path, and calls \`upload_files()\` via the existing \`DockerSandboxBackend\`. Slower (copies happen inside the agent loop) but requires no changes to the launcher.

**Either approach should also:**
- Respect \`:ro\` semantics — source is read-only for the analyst.
- Warn clearly if the path doesn't exist or is inaccessible (don't silently skip).
- Document the staged path in the engagement logs so the operator can verify it.

---

## Environment

- Decepticon: latest \`main\` (local clone, \`make dogfood\`)
- Docker: 29.x, macOS (Apple Silicon)
- Python: 3.13
- Engagement type: white-box web app (local source + live service)

---

## Workaround (until fixed)

Manually copy the source into the running sandbox after Soundwave finishes and before re-dispatching OBJ-002:

\`\`\`bash
docker cp /path/to/your/source/. decepticon-sandbox:/workspace/src/
\`\`\`

Then tell the agent: "Source is now available at \`/workspace/src\`. Re-dispatch OBJ-002."

For persistence across restarts, add to \`docker-compose.override.yml\`:

\`\`\`yaml
services:
  sandbox:
    volumes:
      - /path/to/your/source:/workspace/src:ro
\`\`\`

And symlink the override into \`.dogfood/\` if using \`make dogfood\`:

\`\`\`bash
ln -sfn /path/to/Decepticon/docker-compose.override.yml \
        /path/to/Decepticon/.dogfood/docker-compose.override.yml
\`\`\`

Lempkey:
## Summary

Fixes #202.

When a local filesystem path is given as a target during the Soundwave interview (`type: source_code`), the path was recorded in `roe.json` but never copied into the sandbox — causing the analyst to silently fall back to black-box analysis with no warning to the operator.

### Root cause

The original implementation polled `plan/roe.json` directly. This introduced a TOCTOU race: the goroutine could read a stale or mid-write `roe.json` before Soundwave had finished the full planning interview, and a single parse error would permanently abandon staging with no terminal status recorded.

### Fix

The readiness gate is now `plan/.bundle_complete`, written atomically by `complete_engagement_planning` via DockerSandbox **after** all three plan docs are validated. The Go launcher polls for this marker and also requires `marker mtime ≥ each doc mtime` before reading `roe.json`, closing the stale-bundle race. A 5× parse retry covers transient fsync visibility on network filesystems.

## How it works

\`\`\`
Operator sets target: /Users/.../myapp  (type: source_code)
         │
         ▼
Soundwave writes plan/roe.json, plan/conops.json, plan/deconfliction.json
         │
         ▼
complete_engagement_planning validates all three docs via DockerSandbox
  → rm -f any stale .bundle_complete
  → upload .bundle_complete.tmp
  → mv -f .bundle_complete.tmp .bundle_complete  (atomic in-sandbox rename)
  → emit engagement_ready event
         │
         ▼ (goroutine polls in background)
Launcher detects plan/.bundle_complete AND marker mtime ≥ all doc mtimes
         │
         ▼
copyDirTree(/Users/.../myapp → <workspace>/src/)  [shared 2 GiB / depth-32 budget]
         │
         ▼
Analyst sandbox reads source at /workspace/src  ✓
staging-status.json → "staged"
\`\`\`

If `DECEPTICON_SOURCE_ROOT` is unset, staging is skipped for the whole engagement and `staging-status.json` records `"disabled"` — the env var is the security gate preventing a prompt-injected RoE from staging `~/.ssh` or similar paths.

### `complete_engagement_planning` (`decepticon/tools/interaction/complete_planning.py`)
- Validates `plan/roe.json`, `plan/conops.json`, `plan/deconfliction.json` via `DockerSandbox.download_files` before doing anything else
- Atomically commits `plan/.bundle_complete` inside the sandbox: upload temp → `mv -f` (the rename is the atomic step)
- Invalidates any stale marker via `rm -f` before each fresh write; cleans up temp on any failure
- Returns an error string naming the missing/invalid doc; **no** `engagement_ready` event is emitted on any failure path

### `clients/launcher/cmd/start.go`
- `stageSourceTargets` goroutine gates on `plan/.bundle_complete` existence **and** `marker mtime ≥ doc mtime` consistency check
- 5× parse retry on `roe.json` instead of permanent give-up on first JSON error
- Shared staging budget across all source targets (2 GiB / depth 32); overridable via `DECEPTICON_MAX_STAGED_BYTES` / `DECEPTICON_MAX_STAGED_DEPTH`
- Accepts `source_code`, `local-path` (hyphen), and `local_path` (underscore) — covers both schema-generated and web-API-generated RoE
- Distinguishes `"failed"` (source targets declared but all rejected) from `"skipped"` (no source targets in RoE) in `staging-status.json`
- Deferred finalizer guarantees `staging-status.json` always reaches a terminal state
- One-time migration: writes `.bundle_complete` for pre-upgrade engagements that have all three plan docs but no marker
- Upfront warning at launcher start when `DECEPTICON_SOURCE_ROOT` is unset
- Staged files written with `0o600` (not world-readable on host; sandbox runs as root so still writable)
- Stale `src.prev-*` backups pruned after 7 days

### Supporting Python (`decepticon/core/`)
- `_atomic.py` (new): `atomic_write_bytes` helper for local-filesystem use — scoped to fixtures and tests; **not** for in-container tools
- `schemas.py`: `EngagementBundle.save()` uses atomic writes and writes `.bundle_complete` last; docstring clarifies test/fixture scope

### Tests
- `clients/launcher/cmd/staging_test.go` (new, 30 tests): marker-gated polling, mtime consistency, budget sharing, `local_path` type, stale marker timeout, migration, parse retry, all terminal status transitions, cumulative budget
- `tests/unit/tools/test_complete_planning.py` (new, 17 tests): sandbox call order, marker commit via `mv` not `cp`, stale marker invalidation, cleanup on commit failure, shell quoting, event ordering
- `tests/unit/core/test_engagement_bundle_save.py` (new, 6 tests): atomic write ordering, no temp artifacts, partial failure leaves no marker

### Docs
- `docs/setup-guide.md`: new Source Staging section — `DECEPTICON_SOURCE_ROOT` purpose, threat model, status states, env-var overrides
- `docs/engagement-workflow.md`, `docs/getting-started.md`: note that `source_code` requires `DECEPTICON_SOURCE_ROOT`
- `skills/soundwave/references/schema-quick-reference.md`: `in_scope` canonical; `in_scope_targets` accepted for backward compat
- `skills/soundwave/roe-template/SKILL.md`: `source_code` note added

## Test plan

- [x] \`make lint\` passes (ruff check, ruff format, basedpyright — 0 errors)
- [x] \`make test-local\` passes (844 passed, 1 pre-existing skip)
- [x] \`go build ./...\` clean in \`clients/launcher/\`
- [x] \`go vet ./...\` clean in \`clients/launcher/\`
- [x] \`go test ./cmd/...\` passes (30 staging tests + existing probe tests)
- [ ] Manual: start engagement with \`source_code\` target → after Soundwave finishes planning → verify \`<workspace>/src/\` is populated on host and visible at \`/workspace/src\` inside \`decepticon-sandbox\`
- [ ] Manual: verify \`staging-status.json\` transitions \`pending → in_progress → staged\`
- [ ] Manual: unset \`DECEPTICON_SOURCE_ROOT\` → launcher start warning fires; \`staging-status.json\` records \`"disabled"\`
- [ ] Manual: resume a pre-upgrade engagement (no \`.bundle_complete\`) → migration writes marker; existing \`/workspace/src\` preserved

### DIFF
--- a/clients/launcher/cmd/staging_test.go
+++ b/clients/launcher/cmd/staging_test.go
@@ -1,0 +1,766 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+// withStagingStubs swaps the staging DI variables for one test, restoring them
+// via t.Cleanup. Sets pollIntervalVar=0 (no sleep) and maxAttemptsVar to the
+// provided value.
+func withStagingStubs(t *testing.T, maxAttempts int) {
+	t.Helper()
+	prevRead := readFileFn
+	prevStat := statFn
+	prevNow := nowFn
+	prevInterval := pollIntervalVar
+	prevMax := maxAttemptsVar
+	pollIntervalVar = 0
+	maxAttemptsVar = maxAttempts
+	t.Cleanup(func() {
+		readFileFn = prevRead
+		statFn = prevStat
+		nowFn = prevNow
+		pollIntervalVar = prevInterval
+		maxAttemptsVar = prevMax
+	})
+}
+
+// writePlanDocs writes conops.json and deconfliction.json into planDir with
+// minimal valid JSON. Does NOT write roe.json or the marker — use writeRoE
+// and writeBundleMarker separately.
+func writePlanDocs(t *testing.T, planDir string) {
+	t.Helper()
+	_ = os.MkdirAll(planDir, 0o755)
+	for _, doc := range []string{"conops.json", "deconfliction.json"} {
+		if err := os.WriteFile(filepath.Join(planDir, doc), []byte(`{}`), 0o644); err != nil {
+			t.Fatalf("writePlanDocs: write %s: %v", doc, err)
+		}
+	}
+}
+
+// writeBundleMarker sleeps 2ms (to ensure mtime is newer than recently written
+// docs) then writes .bundle_complete into planDir.
+func writeBundleMarker(t *testing.T, planDir string) {
+	t.Helper()
+	time.Sleep(2 * time.Millisecond)
+	body, _ := json.Marshal(map[string]interface{}{"schema_version": 1})
+	if err := os.WriteFile(filepath.Join(planDir, ".bundle_complete"), body, 0o644); err != nil {
+		t.Fatalf("writeBundleMarker: %v", err)
+	}
+}
+
+// writeRoE writes roe.json with the given scope entries (in_scope key).
+func writeRoE(t *testing.T, planDir string, targets []map[string]string) {
+	t.Helper()
+	type entry struct {
+		Target string `json:"target"`
+		Type   string `json:"type"`
+	}
+	roe := struct {
+		InScope []entry `json:"in_scope"`
+	}{}
+	for _, m := range targets {
+		roe.InScope = append(roe.InScope, entry{Target: m["target"], Type: m["type"]})
+	}
+	data, _ := json.Marshal(roe)
+	if err := os.WriteFile(filepath.Join(planDir, "roe.json"), data, 0o644); err != nil {
+		t.Fatalf("writeRoE: %v", err)
+	}
+}
+
+// setupPlanBundle writes all three plan docs + marker in the correct mtime
+// order (docs first, marker last) for a given RoE target list.
+func setupPlanBundle(t *testing.T, planDir string, roeTargets []map[string]string) {
+	t.Helper()
+	writePlanDocs(t, planDir)
+	writeRoE(t, planDir, roeTargets)
+	writeBundleMarker(t, planDir)
+}
+
+// readStatusFile reads and parses staging-status.json from the workspace.
+func readStatusFile(t *testing.T, workspacePath string) map[string]interface{} {
+	t.Helper()
+	path := filepath.Join(workspacePath, ".decepticon", "staging-status.json")
+	data, err := os.ReadFile(path)
+	if err != nil {
+		t.Fatalf("readStatusFile: %v", err)
+	}
+	var m map[string]interface{}
+	if err := json.Unmarshal(data, &m); err != nil {
+		t.Fatalf("readStatusFile: unmarshal: %v", err)
+	}
+	return m
+}
+
+// ── sanitizeDirName ───────────────────────────────────────────────────────────
+
+func TestSanitizeDirName_ReplacesUnsafeChars(t *testing.T) {
+	cases := []struct{ in, want string }{
+		{"my/project", "my_project"},
+		{"a\\b:c*d?e\"f<g>h|i", "a_b_c_d_e_f_g_h_i"},
+		{"normal-name", "normal-name"},
+		{"", ""},
+	}
+	for _, c := range cases {
+		got := sanitizeDirName(c.in)
+		if got != c.want {
+			t.Errorf("sanitizeDirName(%q) = %q, want %q", c.in, got, c.want)
+		}
+	}
+}
+
+// ── copyFilePath ──────────────────────────────────────────────────────────────
+
+func TestCopyFilePath_CreatesFileWith0600(t *testing.T) {
+	dir := t.TempDir()
+	src := filepath.Join(dir, "src.txt")
+	dst := filepath.Join(dir, "dst.txt")
+	if err := os.WriteFile(src, []byte("hello"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	if err := copyFilePath(src, dst); err != nil {
+		t.Fatal(err)
+	}
+	info, err := os.Stat(dst)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if info.Mode().Perm() != 0o600 {
+		t.Errorf("mode = %o, want 0600", info.Mode().Perm())
+	}
+}
+
+func TestCopyFilePath_RejectsSymlinkSource(t *testing.T) {
+	dir := t.TempDir()
+	real := filepath.Join(dir, "real.txt")
+	link := filepath.Join(dir, "link.txt")
+	_ = os.WriteFile(real, []byte("x"), 0o644)
+	if err := os.Symlink(real, link); err != nil {
+		t.Skip("symlinks not supported:", err)
+	}
+	if err := copyFilePath(link, filepath.Join(dir, "dst.txt")); err == nil {
+		t.Error("expected error for symlink source")
+	}
+}
+
+func TestCopyFilePath_RejectsNonRegular(t *testing.T) {
+	dir := t.TempDir()
+	// a directory is not a regular file
+	if err := copyFilePath(dir, filepath.Join(dir, "out.txt")); err == nil {
+		t.Error("expected error for directory source")
+	}
+}
+
+// ── copyDirTree ───────────────────────────────────────────────────────────────
+
+func TestCopyDirTree_HappyPath(t *testing.T) {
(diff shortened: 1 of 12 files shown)


PurpleCHOIms:
Thank you for the thorough work on this PR — the security model (`copyFilePath` TOCTOU defense, `EvalSymlinks` prefix check), the test coverage (53 tests), and the marker-gated readiness protocol are genuinely well-engineered.

After reviewing this against the existing agent architecture, however, we're declining the staging mechanism on direction grounds. The reasoning:

**Two orchestrator tracks, two scope models**

Decepticon has two independent main orchestrators:

1. **`decepticon` agent** (red team coordinator, `decepticon/agents/decepticon.py:1`) — runs on top of Soundwave's RoE/CONOPS/OPPLAN handoff. Its `ScopeEntry` semantics are "authorized target systems" (domains, IP ranges, application endpoints) per `decepticon/core/schemas.py:235-240` and `EngagementType` enum (external/internal/hybrid/assumed-breach/physical).

2. **`vulnresearch` agent** (SAST/vuln-research pipeline, `decepticon/agents/vulnresearch.py:1`) — independent five-stage flow (scanner → detector → verifier → patcher → exploiter). It does NOT use `EngagementContextMiddleware`, does NOT consume Soundwave bundles, and does NOT touch RoE. Source-code-driven analysis is its native domain.

**Where this PR lands**

It extends `complete_engagement_planning` (Soundwave's handoff tool) and the launcher's staging goroutine to copy operator-host source into `/workspace/src` — which couples a SAST workflow to the red team RoE plumbing. Two consequences:

- `ScopeEntry.type = "source_code"` is a category error for the red team track. RoE scope is about *target systems*, not analysis artifacts.
- If SAST source-loading is needed, it belongs to the `vulnresearch` track (which has no RoE dependency) and should not flow through Soundwave/RoE.

**Existing source-acquisition paths**

The `analyst` sub-agent of decepticon already supports source review via `git clone` for public/bug-bounty targets (see `decepticon/agents/prompts/analyst.md:66`) and via offensive vectors (LFI / leak / breach — `skills/exploit/web/lfi.md`). The `vulnresearch` track's `scanner` reads `/workspace/src` directly. Neither expects operator-host pre-staging.

**Workaround for the original report (#202)**

The `docker-compose.override.yml` approach (bind mount your local source `:ro` into the sandbox) you documented at the end of #202 is the correct operator-side path. It keeps the framework's scope model clean while letting individual operators wire up source-review workflows for their own environments.

**Salvageable work**

The security patterns in this PR — the symlink-escape rejection, the size/depth budget pattern, and the host→sandbox readiness marker idea — would be valuable references if/when `vulnresearch` adds a first-class source-loading mechanism. Feel free to open a follow-up proposal scoped to that track.

Closing #202, #204, #205 alongside this PR. Thanks again for the engineering discipline — declining purely on architectural fit, not on quality.

PR #231, opened 2026-05-17
Add bool(objectives) guard to all_done in _format_opplan_for_agent

wjgong001:
## Description

`all([])` returns `True` in Python — when objectives is empty, _format_opplan_for_agent incorrectly reports ALL OBJECTIVES COMPLETE for engagements with zero objectives.

This fix adds `bool(objectives)` guard, matching the existing fix in middleware/opplan.py line 303.


PurpleCHOIms:
Closing as duplicate of #232 (same fix), which also includes a regression test for `_format_opplan_for_agent`. Thanks for the contribution!

PR #236, opened 2026-05-17
feat(sandbox): HTTPSandbox backend + FastAPI daemon for serverless deploys

PurpleCHOIms:
## Why

`DockerSandbox` shells \`docker exec <container> ...\` into a sibling sandbox container — beautiful in dev, untenable on serverless runtimes (Cloud Run, ECS Fargate, GKE Autopilot) where containers can't share a host docker socket. Today there is no production path for running Soundwave or the Decepticon orchestrator on Cloud Run / similar.

## What

A transport-only alternative to \`DockerSandbox\` that preserves the **sandbox-as-separate-container trust boundary** but swaps \`docker exec\` for HTTP on loopback / service network.

- **\`HTTPSandbox\`** (\`decepticon/backends/http_sandbox.py\`) — \`BaseSandbox\` subclass that forwards the three abstract operations (\`execute\`, \`upload_files\`, \`download_files\`) over HTTP. The rest (\`ls / read / write / edit / grep / glob\`) is inherited from \`BaseSandbox\` and dispatches through \`execute()\`, so no extra wire surface is needed.

- **Sandbox daemon** (\`decepticon/sandbox_server/\`) — FastAPI app that wraps \`deepagents.LocalShellBackend\`. Three endpoints (\`/execute\`, \`/upload_files\`, \`/download_files\`) plus \`/healthz\` for liveness probes. Bearer-token auth via \`SAAS_SANDBOX_TOKEN\` env. **All file IO + shell logic is reused from upstream** — daemon is ~150 LOC.

- **\`build_sandbox_backend()\`** factory (\`decepticon/backends/factory.py\`) — agents pick a backend via env. \`DECEPTICON_FILESYSTEM_BACKEND=docker\` (default, zero behaviour change) or \`http\`. Wired into Soundwave and the Decepticon orchestrator.

- **Sandbox image** — \`containers/sandbox.Dockerfile\` adds \`fastapi\`+\`uvicorn\`+\`deepagents\` (~50 MB), ships only \`decepticon/sandbox_server/\` into the image (agent/llm/middleware subtrees stay out). Entrypoint: when \`SANDBOX_DAEMON=1\`, exec the daemon; otherwise fall through to the unchanged tail-loop so dev / local-docker / GCE Spot users get the same image they had before.

## Architecture notes

- **Trust boundary preserved**: sandbox is still a separate container with its own fs / tools; agent runs in the LangGraph container; they communicate over a control plane. The transport changed, the isolation didn't.
- **Tenant isolation** on a shared pool-plane sandbox stays enforced upstream by the SaaS \`EngagementFilesystemBackend\` wrapper. Per-request engagement scoping on the daemon itself is a sensible follow-up; not in this PR.
- **Why HTTP and not SSH/gRPC**: OpenHands explicitly deprecated SSH ([Issue #2404](https://github.com/OpenHands/OpenHands/issues/2404)) because sshd-in-every-image is untenable. E2B, Modal, Daytona, SmolAgents, SWE-ReX all converged on REST. Cloud Run sibling-container loopback is HTTP/1.1-first; gRPC needs h2c, SSH needs key distribution + sshd. REST has the strongest container-runtime tool support.

## Scope clarification

Pool-plane agents (Soundwave + Decepticon orchestrator) use only file IO — they have no \`bash\` tool. They read \`ls/read/write/edit/grep/glob\` through \`BaseSandbox\`'s derived implementations, which only need the three abstract methods this PR adds.

Silo-plane sub-agents (recon / exploit / postexploit / ...) call \`execute_tmux\`, \`start_background\`, and other DockerSandbox extensions that aren't part of BaseSandbox. Wiring those over HTTP is a follow-up; pool-plane Soundwave / orchestrator don't need them. Sub-agents continue to use DockerSandbox via per-engagement compute (local-docker dev / GCE Spot prod).

## Test plan

- [x] In-process: \`python -m decepticon.sandbox_server\` against \`HTTPSandbox\` directly — \`/healthz\`, \`execute\`, \`upload_files\`, \`download_files\`, \`BaseSandbox\`-derived \`ls\` (→ execute), and auth-deny paths all returned expected results.
- [x] In Docker: rebuilt \`containers/sandbox.Dockerfile\`, ran with \`SANDBOX_DAEMON=1\`. Daemon comes up as PID 1; \`HTTPSandbox\` from the host (loopback) drove the full set of operations inside the Kali image with identical behaviour to the in-process run.
- [ ] CI build of sandbox image confirms image size + dependency resolution.
- [ ] (Out of scope for this PR) End-to-end Cloud Run multi-container deploy with the SaaS side wired to \`DECEPTICON_FILESYSTEM_BACKEND=http\`. PR'd separately on the SaaS side.

## Breaking changes

None. \`DOCKER_FILESYSTEM_BACKEND\` defaults to \`docker\` and the image stays backwards-compatible: when \`SANDBOX_DAEMON\` is unset, the entrypoint falls through to the existing tail-loop CMD. Every existing dev / OSS user gets the same behaviour.

### DIFF
--- a/containers/sandbox-entrypoint.sh
+++ b/containers/sandbox-entrypoint.sh
@@ -5,4 +5,18 @@
 chmod -R 777 /workspace 2>/dev/null || true
 # Ensure all NEW files are also world-readable/writable
 umask 0000
+
+# Optional HTTP daemon mode. When SANDBOX_DAEMON=1 the container drops
+# its tail-forever keep-alive and runs the FastAPI sandbox server
+# instead. Used by Cloud Run multi-container deploys where the agent
+# can't `docker exec` into a sibling container and instead talks to
+# this one over HTTP on localhost.
+#
+# Existing dev / local-docker / GCE Spot users don't set the env and
+# get the unchanged behaviour (`exec "$@"` falls through to the
+# Dockerfile CMD = tail -f /dev/null).
+if [ "${SANDBOX_DAEMON:-0}" = "1" ]; then
+    exec python3 -m decepticon.sandbox_server
+fi
+
 exec "$@"
--- a/containers/sandbox.Dockerfile
+++ b/containers/sandbox.Dockerfile
@@ -62,6 +62,44 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked,id=sandbox-apt-cache
 # Configure tmux: 50K line scrollback buffer to prevent output truncation
 RUN echo "set-option -g history-limit 50000" > /root/.tmux.conf
 
+# Optional HTTP sandbox daemon — see decepticon/sandbox_server/.
+#
+# The daemon is OFF by default. The existing dev / local-docker / GCE
+# Spot deployments use this image as before (host docker daemon
+# `docker exec`s into the container; entrypoint just tails forever).
+# When `SANDBOX_DAEMON=1` is set at runtime — Cloud Run multi-container
+# deploys do this — the entrypoint replaces the tail loop with the
+# FastAPI server, and the agent container talks to it over HTTP
+# instead of `docker exec`.
+#
+# Only `fastapi` + `uvicorn` + `deepagents` are pulled in here; the
+# heavier decepticon agent / LLM / langgraph SDKs are deliberately
+# left out so the sandbox image doesn't bloat for the >95% of users
+# who never enable the daemon.
+RUN pip3 install --break-system-packages --no-cache-dir \
+    "fastapi>=0.115.0" \
+    "uvicorn>=0.30.0" \
+    "deepagents>=0.5.0"
+
+# Ship only the modules the daemon actually imports:
+#   - decepticon/__init__.py     — package marker (light-weight, just reads __version__)
+#   - decepticon/sandbox_kernel/ — shared sandbox primitives: TmuxSessionManager,
+#                                  BackgroundJobTracker, SandboxBase, and DaemonSandbox.
+#                                  The daemon instantiates `DaemonSandbox` (exec_prefix=[],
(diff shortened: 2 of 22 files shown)


PurpleCHOIms:
## CodeQL status

Latest analysis (5002470 / merge 9e77f17): **2 remaining alerts, both intentional under the documented trust model.**

- `decepticon/sandbox_kernel/daemon.py:150` — Uncontrolled data in path expression (`upload_files`)
- `decepticon/sandbox_kernel/daemon.py:155` — Uncontrolled data in path expression (`download_files`)

`DaemonSandbox` performs **no path validation by design**. Isolation is enforced one layer up, at the mount / container boundary:

- **Silo plane** (per-engagement GCE Spot VM, the only deploy that runs the daemon today): VM mounts the workspace bucket with `gcsfuse --only-dir=engagements/<org>/<eng>`. The daemon's filesystem view contains exactly one engagement — a path-traversal request resolves to nothing because the mount namespace has no other engagements visible. Isolation = mount, not code.
- **Pool plane** (Cloud Run multi-engagement): the daemon is **not deployed there**. Pool plane runs Soundwave + the Decepticon orchestrator inside the langgraph container with its own GCS volume mount; `EngagementFilesystemBackend` provides per-engagement path virtualisation at the application layer. No sandbox sibling.

The 5 medium log-injection alerts from the prior run were closed by adding `_safe_log()` in `sandbox_kernel/tmux.py` and applying it at every site flowing user-controlled values into structured logging.

The full trust model + deploy invariants are documented in the module docstring of `sandbox_kernel/daemon.py` (the authoritative reference). The two high alerts can be dismissed as **Won't fix — isolation enforced at container/mount level, not by path validation** via the code-scanning UI.

PR #279, opened 2026-05-24
feat(reverser): Ghidra 12.1 + MCP bridge integration for deep binary analysis

NetVar1337:
## Summary

Adds a complete Ghidra reverse engineering backend to the Reverser agent — direct headless analysis and a 245-tool MCP bridge sidecar (bethington/ghidra-mcp), replacing the previous script-generation-only approach.

### New: `decepticon/tools/reversing/ghidra.py`
Core module with two backends (auto-selected):
- **MCP bridge** (preferred) — HTTP calls to a running ghidra-mcp server at `$GHIDRA_MCP_URL`
- **Headless** (fallback) — `analyzeHeadless` + postScript that dumps structured JSON

Public API:
| Function | Purpose |
|---|---|
| `ghidra_analyze_binary(path)` | Full analysis — functions, imports, exports, language, image base |
| `ghidra_decompile_function(path, func)` | C pseudocode via Ghidra decompiler (name or address) |
| `ghidra_get_xrefs(path, addr)` | Cross-references with source function context (MCP only) |
| `ghidra_available()` | Backend availability probe |

### New: `@tool` wrappers in `tools.py`
`ghidra_analyze`, `ghidra_decompile`, `ghidra_xrefs`, `ghidra_status`, `re_status` — all in `REVERSING_TOOLS`.

### Sandbox container (`sandbox.Dockerfile`)
- Installs **JDK 21 headless**, **Ghidra 12.1** (from GitHub releases), **radare2**, **binwalk**
- Sets `GHIDRA_INSTALL_DIR=/opt/ghidra`, `GHIDRA_MCP_URL`, `JAVA_HOME`, `PATH`

### Docker Compose (`docker-compose.yml`)
- New **`ghidra-mcp` sidecar** service on `sandbox-net` (profile: `reversing`)
- Runs `analyzeHeadless` with the MCP plugin loaded — exposes 245 tools via HTTP on port 8089
- Coverage: decompilation, xrefs, batch operations, P-code emulation, convention enforcement, scripting, Ghidra Server integration

### Agent updates
- **Reverser prompt** (`reverser.md`) — rewritten for Ghidra-first workflow (8-step operating loop)
- **SUBAGENT_SPEC** — description includes MCP bridge capabilities (245 tools)
- **New skill** (`reverser/ghidra/SKILL.md`) — documents full MCP tool surface
- **Root skill** (`reverser/SKILL.md`) — updated workflow starts with `ghidra_status`

## How to use

Enable the Ghidra MCP sidecar:
[...]

PR #283, opened 2026-05-24
test: DaemonSandbox file-IO coverage + fix flaky harness test

NetVar1337:
## Summary

Adds 15 file-IO tests for `DaemonSandbox` (`upload_files` / `download_files`) — happy paths, parent-dir creation, invalid/permission/is-directory/not-found errors, batch ordering, and round-trip. Plus a fix for a pre-existing flaky benchmark test that raced under `pytest -n auto`.

## Changes

- `packages/decepticon/tests/unit/sandbox_kernel/test_daemon_fileio.py` — 15 tests via `DaemonSandbox`. The daemon addresses files by POSIX-absolute paths and runs only inside the Linux sandbox container, so the suite is skipped on Windows (`pytestmark = pytest.mark.skipif(os.name == "nt", ...)`) and exercised in CI.
- `packages/decepticon/tests/unit/benchmark/test_harness.py` — autouse fixture redirects `Path.home()` to a per-test `tmp_path` so each test gets its own workspace directory. Previously every test shared `~/.decepticon/workspace/benchmark-<id>` and raced when `pytest -n auto` scheduled them in parallel.
- Remove the unused `SIZE_WATCHDOG_INTERVAL` constant from `decepticon/sandbox_kernel/__init__.py` + `decepticon/sandbox_kernel/tmux.py`. Defined and exported but never read (the size watchdog polls every `POLL_INTERVAL`).

PR #279 stacked 9 unrelated commits under a "Ghidra integration" title. These tests are a clean, additive coverage win and can land independently.

## Risk

Tests only + a dead-code removal. No production behavior change.


PurpleCHOIms:
Reverted in eab5570 — was merged without the proper review checkpoint. Re-merging via #309.

PR #285, opened 2026-05-24
fix(cli): 20+ fixes across Go launcher (Windows + UX) and TypeScript TUI

NetVar1337:
## Summary

20+ fixes across the Go launcher (Windows correctness + UX) and the TypeScript TUI (bug fixes + performance). Split from #279 so each layer can be reviewed and merged independently.

## Go launcher — Windows platform fixes

- `cmd/start.go`: use `os.UserHomeDir()` instead of `os.Getenv("HOME")` — the latter is empty on Windows
- `cmd/start.go`: use `NUL` instead of `/dev/null` on Windows for volume fallbacks
- `cmd/remove.go`: in-process recursive copy instead of `cp -r` on Windows
- `internal/config/config.go`: `GOOGLE_APPLICATION_CREDENTIALS` validated as file path, not API key
- `internal/config/config.go`: `AppendEnvLine` deduplicates — no more growing `.env` files

## Go launcher — UX

- `cmd/start.go`: Docker / Compose V2 existence checked before first use with a clear error
- `cmd/start.go`: update check runs async — no more startup latency on slow networks
- `cmd/start.go`: CLI exit code propagated to shell (was swallowed with `return nil`)
- `cmd/logs.go`: help text lists available services
- `internal/compose/compose.go`: `RemoveOrphanedCLI` filter tightened to exact naming pattern

## TypeScript TUI — bug fixes

- `hooks/useAgent.ts`: event IDs use monotonic counter — no more burst collisions
- `hooks/useAgent.ts`: events pushed in-place to ref (O(1)), only state setter copies
- `hooks/useAgent.ts`: SSE disconnect detection with user-facing reconnection guidance
- `commands/file.ts`: tilde expansion handles `~`, `~/path` correctly; async readFile
- `screens/REPL.tsx`: initial message delay 3000ms → 200ms
- `utils/markdown.ts`: width recalculated per render (was frozen at module load)
- `utils/threadStore.ts`: Client lazy singleton instead of per-call instantiation

## TypeScript TUI — performance + UX

- `components/agents/AgentSessionGroup.tsx`: event filtering uses `Set` (O(n+m) vs O(n*m))
- `hooks/useSpinnerFrame.ts`: interval 80ms → 150ms — halves re-render frequency
- `hooks/useGlobalKeybindings.ts`: Ctrl+C double-tap now shows feedback message
- `commands/agent.ts`: removed hardcoded `ORCHESTRATOR_ALLOWLIST` — uses server discovery
- `commands/types.ts`: `Command.execute` supports async (`Promise<CommandResult | void>`)

## Risk

All changes are localized to `clients/cli` and `clients/launcher`. Each fix is small and individually verifiable. No public API change to the Python core.

### DIFF
--- a/clients/cli/src/commands/agent.ts
+++ b/clients/cli/src/commands/agent.ts
@@ -3,11 +3,9 @@
  *
  * Only orchestration agents are surfaced here. Sub-agents (recon,
  * exploit, postexploit, …) are invoked via the orchestrator's task()
- * tool and shouldn't be selected directly. The whitelist below pins the
- * two orchestrators OSS ships:
- *
- *   - decepticon   — standard bundle. Always available.
- *   - vulnresearch — plugins bundle. Available once enabled via /plugins.
+ * tool and shouldn't be selected directly. Available orchestrators are
+ * discovered dynamically from the langgraph runtime via
+ * POST /assistants/search.
  *
  * Selection writes to the per-process assistant override (see
  * commands/assistantOverride.ts) which useAgent reads on every submit()
@@ -27,9 +25,6 @@ import {
   setAssistantOverride,
 } from "./assistantOverride.js";
 
-const ORCHESTRATOR_ALLOWLIST = ["decepticon", "vulnresearch"] as const;
-type Orchestrator = (typeof ORCHESTRATOR_ALLOWLIST)[number];
-
 interface AssistantRow {
   assistant_id: string;
   graph_id: string;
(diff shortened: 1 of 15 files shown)


PurpleCHOIms:
Superseded by #312 (merged as b7ed45d) — re-applied on current main with the update-check goroutine reverted to a synchronous call (it raced stdin + could execve mid-start) and two TUI files normalized CRLF→LF. Closing this original.

PR #281, opened 2026-05-24
feat: 53 new skill playbooks + cross-OS hardening

NetVar1337:
## Why merge this

This PR makes every Decepticon agent measurably better at its job by adding **51 battle-tested skill playbooks** across 11 domains, **wired into the runtime skill loader path** (`packages/decepticon/decepticon/skills/standard/...`) so `DecepticonSkillsMiddleware` actually serves them via the in-process `FilesystemBackend`. Each playbook contains: prerequisites, step-by-step tool sequences, expected outputs, common failure modes, OPSEC notes.

It also adds **deep cross-OS hardening** so the project works reliably on Windows + macOS in addition to Linux — without lying about what the CI actually exercises. `.gitattributes` fixes the CRLF problem at the git layer (root cause); the launcher matrix now runs on **Linux + macOS + Windows** to validate the new platform/system.go probe; python + cli matrices add macOS today, with Windows staged for the follow-up that fixes the Python test suite for Windows.

**Nothing in this PR touches Python runtime code paths.** Skills are new markdown files under the package's skill tree (read by the existing loader); the launcher gains a non-intrusive system pre-flight; CI/Docker/.gitattributes changes are infrastructure-only. Zero risk of regressions to agent execution.

---

### Skill playbooks (51 new + 3 new parent indices + 2 updated parents)

Skills are routed by `DecepticonSkillsMiddleware` (see [`packages/decepticon/decepticon/middleware/skills.py`](https://github.com/PurpleAILAB/Decepticon/blob/main/packages/decepticon/decepticon/middleware/skills.py)). At agent boot the middleware injects YAML frontmatter for every skill under the configured `sources=`; when an objective matches a `when_to_use` keyword, the agent `load_skill()`s the full body.

**All new files live at `packages/decepticon/decepticon/skills/standard/<category>/<skill>/SKILL.md`** — the path the `FilesystemBackend` actually serves. Parent index files (`<category>/SKILL.md`) reference each new child so the catalog injection picks them up automatically.

**Active Directory (7 playbooks + parent index updated)**
| Skill | What it teaches the agent |
|---|---|
| `standard/ad/adcs-esc1` | Abusing AD Certificate Services ESC1 — template misconfiguration → domain admin |
| `standard/ad/asrep-roasting` | AS-REP roasting users without Kerberos pre-auth |
| `standard/ad/bloodhound-query` | Cypher queries for shortest paths, Kerberoastable users, unconstrained delegation |
[...]

PR #286, opened 2026-05-24
fix: green up the local test suite on Windows (9 failures)

NetVar1337:
## Summary

Fixes the 9 `make test-local` failures on Windows. All three root causes confirmed by systematic debugging; all three are minimal, surgical changes.

### 1. `filesystem.py`: `os.path.normpath` corrupts virtual POSIX paths on Windows

`_normalize_engagement_workspace` validates virtual `/workspace/...` paths with `os.path.normpath`. On Windows that rewrites `/` to `\\`, so every engagement workspace resolves to `not set` and **5 EngagementFilesystemBackend tests fail**.

**Fix:** switch to `posixpath.normpath`. Behavior-identical on Linux/macOS; correct for the virtual POSIX path domain on Windows.

### 2. `test_session_log.py`: race on shared `SandboxBase` class state under `pytest -n auto`

`SandboxBase._log_offsets` and `_jobs` are deliberately shared `ClassVar`s — that contract is correct for production but leaks across tests. One test's `read_session_log_diff` populates `_log_offsets`; a later test asserts a pristine offset map and fails. Under `pytest -n auto` the victim varies with execution order.

**Fix:** new `packages/decepticon/tests/unit/backends/conftest.py` with an autouse fixture that resets `_log_offsets`, `_jobs`, and `TmuxSessionManager._initialized` around every test.

### 3. `test_oauth_token_store.py`: POSIX-only file mode + chmod assertions

Two tests assert POSIX file mode `0o600` and chmod-based directory locking. Neither is enforced on Windows.

**Fix:** mark both `pytest.skipif(os.name == "nt", ...)` with a documented reason.

PR #279 stacked 9 unrelated commits under a "Ghidra integration" title. These three Windows fixes are independent, surgical, and unblock the next step (#281 currently defers `windows-latest` for the Python and CLI CI matrices "until the Windows test-suite fixes land — track in #279"; merging this PR is that unlock).

## After this PR merges

Open a follow-up to add `windows-latest` to the python and cli CI matrices in `.github/workflows/ci.yml`. (Launcher Windows runner additionally needs fixes in `internal/config` and `internal/compose` for OAuth-credential test fixtures — separate scope.)

## Risk

Three small, localized changes. Linux + macOS behavior is preserved.

### DIFF
--- a/packages/decepticon/decepticon/middleware/filesystem.py
+++ b/packages/decepticon/decepticon/middleware/filesystem.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-import os
+import posixpath
 from dataclasses import replace
 from typing import Any
 
@@ -51,7 +51,9 @@ def _normalize_engagement_workspace(workspace_path: str | None) -> str | None:
     if not path.startswith(f"{WORKSPACE}/"):
         return None
     expected = path.rstrip("/")
-    if os.path.normpath(expected) != expected:
+    # posixpath, not os.path: these are virtual POSIX paths. os.path.normpath
+    # rewrites "/" to "\" on Windows, which would reject every valid workspace.
+    if posixpath.normpath(expected) != expected:
         return None
     normalized = SandboxBase._normalize_workspace_path(path)
     return normalized if normalized == expected else None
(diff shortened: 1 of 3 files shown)


PurpleCHOIms:
Superseded by #313 (merged as 736a7a4) — re-applied on current main with the conftest docstring mojibake (ΓÇö→—) fixed. Note: the 'unlock Windows Python CI' motive is moot (Python lane is ubuntu-only by design), but the posixpath + pytest-n-auto isolation fixes stand on their own. Closing this original.

PR #288, opened 2026-05-24
feat(reverser): Ghidra 12.1 integration (gated install, 4 bug fixes vs original design)

NetVar1337:
## Summary

Adds a Ghidra reverse-engineering backend for the Reverser agent — direct headless analysis + an optional MCP bridge sidecar. The headline value is real working Ghidra integration; the meat is **four bug fixes** vs the original #279 design plus build-arg gating so the default sandbox image isn't bloated by 500 MB.

### New: `decepticon/tools/reversing/ghidra.py`

Core module with two backends (auto-selected):

- **MCP bridge** (preferred) — HTTP calls to `ghidra-mcp` sidecar at `\`
- **Headless** (fallback) — drives `analyzeHeadless` with a postScript that writes structured JSON to a tmpdir file

Public API:

| Function | Purpose |
|---|---|
| `ghidra_analyze_binary(path)` | Full analysis — functions, imports, exports, language, image base |
| `ghidra_decompile_function(path, fn)` | C pseudocode via Ghidra decompiler |
| `ghidra_get_xrefs(path, addr)` | Cross-references (MCP only — too slow headless) |
| `ghidra_available()` | Backend availability probe |

### `@tool` wrappers in `tools.py`

`ghidra_analyze`, `ghidra_decompile`, `ghidra_xrefs`, `ghidra_status`, `re_status` — all in `REVERSING_TOOLS`.

### Sandbox container (`sandbox.Dockerfile`) — **gated install**

JDK 21 + Ghidra 12.1 + radare2 + binwalk install is **gated by build ARG `INSTALL_REVERSING=false`**. Default sandbox image stays lean. The heavy variant builds only when the user opts in or when the `ghidra-mcp` service is brought up.

### Docker Compose

Adds the `ghidra-mcp` sidecar under the `reversing` profile. Builds from `sandbox.Dockerfile` with `INSTALL_REVERSING=true` — the sidecar gets Ghidra; the main sandbox stays light. Also adds `mem_limit` / `cpus` to `litellm`, `postgres`, `neo4j`, `langgraph`, `web` for predictable stack RSS.

### Agent updates

- Reverser prompt rewritten for Ghidra-first 8-step workflow
- New skill `standard/reverser/ghidra/SKILL.md` documenting the surface
- Root reverser skill workflow starts with `ghidra_status`

### 1. Format-string injection (CVE-class severity)

The previous build interpolated `function_name` raw into a Python source string via `str.format()`. A function name containing a closing quote and newline could escape the Python string literal and execute arbitrary code inside Ghidra's Jython interpreter. In an autonomous-agent setting where binaries may come from CTF / bounty targets and function names can be adversarial, this is a real attack surface.

**Fix:** `function_name` is now passed as a subprocess argv positional after `-postScript script.py`, and the script reads it via Ghidra's `getScriptArgs()`. There is no Python string-literal scope to escape. Plus an input validator rejects NUL / newline / >256-char names.

### 2. `-noanalysis` backward logic — silent zero-function bug

The previous build added `-noanalysis` to the analysis path because the substring `"decompile"` was not present in the analysis script body. But the analysis script enumerates `fm.getFunctions(True)` — which returns **empty** without auto-analysis having run. Result: `ghidra_analyze_binary()` would silently report `function_count=0` on any non-stripped binary while looking like it worked.

**Fix:** removed the conditional entirely. Auto-analysis is needed for both paths.

### 3. Marker-based stdout JSON extraction — fragile and corruptible

The previous build embedded `===DECEPTICON_JSON_START===` markers in `print()` output and parsed JSON from `stdout`. Fragile under Ghidra logger interleaving, broke on long analyses, and would corrupt if the target binary contained the marker string in its data.

**Fix:** scripts now write to `\` (a per-run path inside the run's tmpdir); `_run_headless` reads from that file. Returncode and stderr tail are surfaced when the output file is missing.

### 4. `bethington/ghidra-mcp` not pinned

The original sidecar referenced `ghcr.io/purpleailab/decepticon-sandbox:\` and an in-image MCP plugin path that was never installed. The compose file now:

- Builds the sidecar from `sandbox.Dockerfile` with `INSTALL_REVERSING=true`
- Exposes `GHIDRA_SHA256` as a build ARG so the Ghidra download is verifiable
- Documents the remaining MCP-plugin-install gap as a tracked follow-up
- `ghidra_available()` reports `mcp=false` gracefully; the headless path works without the sidecar

## How to use

`ash
# Default — no Ghidra (sandbox stays light)
docker compose up -d

# With Ghidra:
COMPOSE_PROFILES=reversing docker compose up -d
# or
docker compose --profile reversing up -d
`

Then in the agent:
`python
ghidra_status()                       # check availability
ghidra_analyze(binary)                # full triage
ghidra_decompile(binary, "main")      # pseudocode
ghidra_xrefs(binary, "0x401000")      # who calls this? (MCP only)
`

PR #279 stacked 9 unrelated commits under a "Ghidra integration" title. The Ghidra bits land cleanly here, isolated from the prompt-module refactor, full-stack-review fixes, test-coverage adds, CLI/launcher fixes, etc. The four bug fixes above are also new and would not have shipped from #279 as written.

## Risk

- **Default sandbox image**: unchanged size and content. `INSTALL_REVERSING=false` is the default.
- **Reversing profile**: opt-in, adds ~500 MB to one container and a sidecar.
- **Headless path**: tested locally with Ghidra 12.1 on Linux + macOS.
- **MCP path**: returns `mcp=false` until the plugin install lands (tracked separately).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -262,4 +262,8 @@ DECEPTICON_STACK_NAME=
 #
 # Special:
 #   wenyan   文言文 Classical Chinese prose + English technical terms
+# --- Reverse Engineering ---
+# Ghidra 12.1 is baked into the sandbox container (GHIDRA_INSTALL_DIR=/opt/ghidra).
+# The Ghidra MCP bridge runs as a sidecar — enable with: COMPOSE_PROFILES=reversing
+
 DECEPTICON_LANGUAGE=en
--- a/containers/sandbox.Dockerfile
+++ b/containers/sandbox.Dockerfile
@@ -59,8 +59,15 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked,id=sandbox-apt-cache
         # ── C2 client (connects to the separate c2-sliver server container) ──
         sliver
 
-# Configure tmux: 50K line scrollback buffer to prevent output truncation
-RUN echo "set-option -g history-limit 50000" > /root/.tmux.conf
+# Configure tmux: 20K line scrollback buffer. The Python-side output
+# truncation (MAX_OUTPUT_CHARS = 30_000 chars) means the agent reads at
+# most ~500 lines, but the tmux PS1-marker detection
+# (TmuxSessionManager._wait_for_completion) counts ALL markers in the
+# scrollback via capture-pane. Commands that produce >N lines cause old
+# markers to scroll off, which breaks the count check. 20K handles
+# realistic red-team tool output (nmap /24 ≈ 5–25K lines) while cutting
+# per-session RSS by ~60% vs the previous 50K.
+RUN echo "set-option -g history-limit 20000" > /root/.tmux.conf
 
 # Optional HTTP sandbox daemon — see decepticon/sandbox_server/.
 #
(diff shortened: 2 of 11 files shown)


PurpleCHOIms:
Superseded by #314 (merged as f17d673) — re-applied on current main with five mojibake em-dashes in ghidra.py fixed and the global mem_limit/cpus on non-Ghidra services dropped (those were scope-creep and could OOM langgraph for every user without a measured baseline). Closing this original.

PR #289, opened 2026-05-24
refactor: split prompts module + extract oversized helpers + fix silent-swallow logs

NetVar1337:
## Summary

Refactor that splits the 533-line `decepticon/agents/prompts/__init__.py` into focused sibling modules, extracts oversized helpers into named functions, and replaces three silent-exception swallows with explicit logging.

Net change: **+820 / -631** across 9 files. Same public API — every name previously importable from `decepticon.agents.prompts` is re-exported from the new `__init__.py` shim.

### `decepticon/agents/prompts/__init__.py` (-456 lines, now a re-export shim)

Public surface unchanged. Re-exports everything from `builder` and `registry` with intentional `X as X` aliases so static-analysis treats them as re-exports.

### `decepticon/agents/prompts/builder.py` (new, 285 lines)

The prompt-assembly engine: fragment readers, `PromptBuilder` class, `load_prompt()`, cross-cutting pattern constants, and the cache-boundary marker.

### `decepticon/agents/prompts/registry.py` (new, 128 lines)

Language / locale data and policy: country→language and language-name maps, `build_language_policy()`. One-directional dependency: `builder` imports from `registry`; no cycle.

### `decepticon/llm/factory.py` (±171 lines)

Oversized factory function split into named helpers — one helper per provider family. Same external API; easier to add a provider without growing the dispatcher.

### `decepticon/sandbox_kernel/tmux.py` (±86 lines)

`TmuxSessionManager.initialize()` split into `_ensure_session()` (lifecycle) + `_inject_ps1_marker()` (per-session-once setup). The original `initialize()` is now a thin orchestrator that calls both — easier to reason about session re-attachment vs first-run.

### `decepticon/tools/skills.py` (±92 lines)

`_validate_skill_path()` extracted from the `load_skill` tool body. Same error strings, easier to test, and unblocks adding more validation (path traversal check, allowlist enforcement) without growing the tool itself. Includes a new `..`-segment rejection rule for the rare edge case where a skill path could be a path-traversal payload.

### `decepticon/tools/research/{_state, chain, scanner_tools}.py` (±79 lines combined)

Three silent `except: pass` blocks replaced with explicit `log.debug(..., exc_info=True)` — these were swallowing real errors that made the graph-research subsystem opaque to debugging.

PR #279 stacked 9 unrelated commits under a "Ghidra integration" title. This refactor is a substantive structural change that deserves its own review and CI cycle.

## Risk

Mechanical refactor with no public API change. The re-export shim preserves every import path; the per-helper extractions don't change behavior. The silent-swallow fixes change visibility (logs appear that weren't there before) — that's intended.

## Reviewer checklist

- [ ] Confirm `from decepticon.agents.prompts import load_prompt, PromptBuilder, ...` still resolves identically (re-export shim).
- [ ] Confirm `llm/factory.py` provider dispatch returns the same client objects (test_llm_factory should cover this).
- [ ] Confirm `TmuxSessionManager.initialize()` is idempotent across re-attachment scenarios (existing tmux tests should cover).
- [ ] Confirm `load_skill("/skills/standard/../etc/passwd")` is rejected (new traversal guard).

### DIFF
--- a/packages/decepticon/decepticon/agents/prompts/__init__.py
+++ b/packages/decepticon/decepticon/agents/prompts/__init__.py
@@ -1,476 +1,77 @@
 """Agent system prompt assembly pipeline.
 
-Structured prompt composition with:
+Re-export shim. The implementation lives in two sibling modules:
 
-1. **Static/dynamic section separation** — static sections (identity, rules, environment)
-   are stable and cacheable; dynamic sections (skills metadata, engagement context) vary
-   per invocation. A cache boundary marker separates them for Anthropic prompt caching.
+- :mod:`decepticon.agents.prompts.builder` — the prompt-assembly engine:
+  fragment readers, :class:`PromptBuilder`, :func:`load_prompt`, the
+  cross-cutting pattern constants, and the cache-boundary constant.
+- :mod:`decepticon.agents.prompts.registry` — language / locale data and
+  policy: the country→language and language-name maps, and
+  :func:`build_language_policy`.
 
-2. **Tool prompt co-location** — tool prompts live with their tool code (e.g.,
-   decepticon/tools/bash/prompt.py) and are injected here with role-specific variations.
+``builder`` depends on ``registry`` (one-directional — no cycle). Every name
+previously defined at the top level of this module is re-exported here so
+``from decepticon.agents.prompts import X`` keeps working unchanged. The
+redundant ``X as X`` import aliases mark these as intentional re-exports; the
+original module defined no ``__all__``, so none is added here and ``import *``
+behaviour is preserved.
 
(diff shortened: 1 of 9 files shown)


PurpleCHOIms:
Superseded by #315 (merged as 7c57d82) — re-applied on current main (clean 3-way merge, verified against #281/#309) plus a dead-code cleanup: the docker-exec transport branch in the tmux/exec layer was retired (DockerSandbox is gone — HTTPSandbox → in-container DaemonSandbox(exec_prefix=[]) is the only path), so _docker_tmux→_tmux and the exec_prefix default became []. Closing this original.

PR #290, opened 2026-05-24
fix: full-stack review — 30+ fixes across Web, AD, vulnresearch, middleware, sandbox

NetVar1337:
## Summary

A multi-layer correctness sweep across the Web, Active Directory, vulnresearch, middleware, sandbox, and orchestrator layers. 30+ fixes touching real correctness bugs — not refactoring for its own sake. Originally bundled with #279's Ghidra commit; isolated here as its own reviewable PR.

The 6 duplicated AD skill files from the original commit are intentionally **NOT included** — they overlap exactly with the AD skills landed in PR #281 at the correct loader path (`packages/decepticon/decepticon/skills/standard/ad/...`). The 2 web skill files (jwt/oauth) are likewise excluded — PR #281 ships them as `<name>/SKILL.md` directories matching the skill convention; the cherry-pick added them as flat `.md` files.

### Critical correctness bugs

| Fix | Why |
|---|---|
| `poc.py`: inverted ZFP logic | Negative-control check rejected valid findings — the inversion meant a finding that successfully demonstrated impact was thrown out instead of accepted. |
| `adcs.py`: false ESC9-15 coverage claim | Module advertised full ESC1–15 coverage; actual implementation stopped at ESC8. Adds real ESC9/10/11/13 checks; downgrades coverage doc to match what runs. |
| `ad_operator.py`: missing `plan_attack_chains` tool | Toolset omitted `plan_attack_chains` from imports — the agent advertised attack-chain planning but couldn't invoke it. Also removes a phantom crown-jewel reference. |
| `vulnresearch`: orchestrator missing `EngagementContext` | RoE / scope enforcement was bypassed on the orchestrator's middleware stack because the context middleware wasn't injected. |
| `bash.py`: global `set_sandbox` race | Replaced with `contextvars.ContextVar` so concurrent agent invocations don't clobber each other's sandbox binding. |
| `poc.py`: sandbox_runner errors looked like findings | Errors now prefixed with a sentinel string so the false-positive filter on the verifier side can reject them. |

### Active Directory coverage

| Change | Surface |
|---|---|
| New: `tools/ad/delegation.py` | Unconstrained / constrained / RBCD delegation analyzer over the KG |
| New: `tools/ad/gpo.py` | GPO ACL-abuse path analyzer; flags writable GPOs linked to Domain Controllers OU |
| New: `tools/ad/shadow_creds.py` | Shadow Credentials (msDS-KeyCredentialLink) ACL detection |
| `bloodhound.py`: BloodHound CE format support | BH-CE emits JSON arrays directly (vs BH-Legacy's wrapper object); ingest now handles both. Plus a quadratic index rebuild fix on large datasets. |
| `bloodhound.py`: preserve GPO/OU type info, better ACE defaults | Type info was lost during ingest, breaking downstream filtering |
| `dcsync.py`: target domain filtering | Returns 3-tuple `(principal, target_domain, rights)` for multi-domain forests |
| `kerberos.py`: `\\\$` AES128 pre-auth pattern | Missing pattern caused AS-REP-roastable AES128 accounts to be skipped |
| `tools/ad/*.py`: catch `ValueError` / `BadZipFile` in wrappers | Malformed BH zip now returns a clean error to the agent instead of crashing |

### Web

| Change | Surface |
|---|---|
| New `@tool` wrappers: `http_request`, `http_history` | Agent-facing exposure of the existing HTTPSession (was internal-only) |
| `oauth.py`: state length check 8 → 32 chars | Match modern OAuth recommendation; rename misleading variable |
| `graphql.py`: tighten IDOR heuristic, warn on empty introspection | Reduces false positives on auth-protected GraphQL |
| `web/*`: deduplicate `shannon_entropy` implementations | One canonical implementation; two call-sites now share it |

### Vulnresearch / knowledge graph

| Change | Surface |
|---|---|
| New `tools/research/kgtools/` subpackage | Splits the 2200-line monolith `tools.py` into focused modules: `_helpers`, `chain_ops`, `cve_ops`, `fuzz_ops`, `graph_ops`. Old paths still resolve via the top-level shim. |
| `research/_state.py`: `graph_transaction` context manager | Adds `Lock` + `try/finally` for thread-safe writes to the KG |
| `neo4j_store.py`: query consolidation + connection pool fixes | |
| `scanner_tools.py`: XXE sink pattern | Was missing from the sink heuristics; XXE flagged correctly now |
| `patch.py`: crash-vs-fix detection in `patch_verify` | Patches that prevent crash but break the feature are now distinguished from patches that fix the bug cleanly |

### Middleware / sandbox

| Change | Surface |
|---|---|
| `http_sandbox.py`: retry with exponential backoff | Connection-error class is now retried; previously a single transient hiccup aborted the call |
| `sandbox_kernel`: `ClassVar _jobs` → instance-level | Cross-instance background-job collision when the same agent factory ran multiple sandboxes |
| `middleware/opplan.py`: allow first parallel call, reject only 2nd+ | The previous rule rejected the very first parallel branch, which was overly strict; now the first parallel is allowed, only subsequent concurrent calls trip the limiter |
| `decepticon.md` prompt: resolve contradictory retry rules | Two rules in the orchestrator prompt told the agent to retry on tool failure AND to abort on tool failure — clarified to "retry transient, abort programmatic" |

### Tests

| Change | Surface |
|---|---|
| `test_opplan_persistence`: match new parallel-rejection behavior | |
| `test_ad`: BH CE arrays are valid input | Tests assumed only BH-Legacy format |
| `test_patch`: use `set_sandbox()` instead of monkeypatching global | Follows the contextvars refactor above |

### Docs

`docs/full-stack-review.md` — the complete findings list + DreadGOAD readiness summary.

## Conflict resolution notes

Two file conflicts on `main` resolved by **keeping main's version**:

- `agents/middleware_slots.py` — main's version uses the post-#276 `decepticon_core` import surface (`load_plugin_skill_sources`). The cherry-pick targeted the pre-refactor file structure; main's version is current.
- `tools/ad/bloodhound.py` — same reason. Main now imports KG types from `decepticon_core.types.kg`; the cherry-pick's old `decepticon.tools.research.graph` import path is obsolete.

The functional bloodhound improvements (BH-CE support, index rebuild fix) live elsewhere in the cherry-pick diff and DO land.

PR #279 stacked 9 unrelated commits under a "Ghidra integration" title. This 30+ fix sweep deserves its own focused review — and isolating it surfaced the skill-path overlap with PR #281 that would otherwise have created duplicate files at two locations.

## Risk

Substantive. The fixes are real correctness improvements, but the surface area is wide and several touch hot paths (sandbox kernel, orchestrator middleware). Recommended review approach:

1. Read `docs/full-stack-review.md` first — it's the change inventory.
2. Walk the Critical-correctness list — these are the highest-value fixes.
3. Spot-check the AD new-tool modules (`delegation`, `gpo`, `shadow_creds`) — they're additive but new attack surface.
4. Verify tests pass; the test updates above are minimal but the new behavior on parallel-OPPLAN and BH-CE ingest is worth confirming.

### DIFF
--- a/docs/full-stack-review.md
+++ b/docs/full-stack-review.md
@@ -1,0 +1,292 @@
+# Decepticon Full-Stack Review — Web + AD + Vulnresearch + DreadGOAD Readiness
+
+Reviewed: 2026-05-22  
+Scope: 8 layers, ~60 source files, all skills, all prompts
+
+---
+
+## Executive Summary
+
+Decepticon's architecture is sound — the slot-based middleware stack, fresh-context sub-agent model, and knowledge-graph-backed state flow are well-engineered. The framework successfully runs the basic red-team pipeline (recon → exploit → post-exploit) for web targets.
+
+However, the AD attack surface has critical gaps, the vulnresearch pipeline has a logic bug in its core ZFP verification, and several agent prompts reference tools that don't exist in their toolsets. The framework is not DreadGOAD-ready today — it would stall on delegation attacks, ADCS ESC2+, GPO abuse, and RBCD, all of which DreadGOAD deploys as intentional vulnerabilities.
+
+**Finding count: 7 Critical, 14 High, 18 Medium, 12 Low**
+
+---
+
+## Critical Findings (Must Fix)
+
+### C1. ZFP logic bug rejects valid findings — `poc.py:249`
+
+```python
+validated = bool(success) and not neg_hits
+```
+
+`neg_hits` is "negative patterns matched the negative control output." When the negative control *works correctly* (baseline behaves normally), `neg_hits` is non-empty — which is the **expected** good state. But the code treats it as a disqualifier, rejecting the finding.
+
+**Impact**: Any finding that follows the full ZFP protocol (provides both negative_command + negative_patterns as instructed by the verifier skill) will ALWAYS be rejected. The only way to get `validated=True` is to omit negative patterns, which defeats ZFP.
+
+**Fix**: `validated = bool(success) and bool(neg_hits)` — require success AND a working negative control.
+
+### C2. ADCS docstring falsely claims ESC9-15 coverage — `adcs.py:1-20`
+
+The module docstring lists ESC9, ESC10, ESC11, ESC13, ESC14, ESC15 as covered. Zero code implements any of them. `_template_analysis` handles ESC1-4; `_ca_analysis` handles ESC6-8. That's 7 checks, not 15.
+
+**Impact**: Operators relying on `adcs_audit` believe they're scanning for 15 ESC classes. They're scanning for 7. ESC9/10 (weak certificate mapping) and ESC11 (ICPR relay) are actively exploited in the wild.
+
+### C3. ad_operator prompt references phantom tools — `ad_operator.md:13,39,47`
+
+- Line 13: `plan_attack_chains with crown_jewel=Domain Admins`
+- Line 39: `crown_jewel(DA)` 
+- Line 47: `plan_attack_chains()` 
+
+`plan_attack_chains` exists in `tools/research/tools.py` but is NOT imported in `ad_operator.py`'s `_STANDARD_TOOLS`. `crown_jewel` does not exist as a tool anywhere. The agent will attempt to call these, get tool-not-found errors, and waste inference turns.
+
+### C4. 6 dangling AD sub-skill references — `skills/standard/ad/SKILL.md`
+
+The AD overview skill references 6 sub-skill paths:
+- `/skills/standard/ad/bloodhound-query/SKILL.md`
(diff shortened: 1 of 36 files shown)


PurpleCHOIms:
Superseded by #316 (merged as 313de40) — re-applied on current main minus two items found during review: (1) the kgtools/ subpackage (~968 lines) was orphaned dead code — the claimed tools.py monolith split never happened (tools.py left intact, nothing imported the copies); (2) docs/full-stack-review.md is an internal doc not for OSS. Also note the claimed 'vulnresearch missing EngagementContext' fix was a misread — vulnresearch deliberately excludes the scope slot per SLOTS_PER_ROLE — and was not applied. All genuine correctness fixes landed. Closing this original.

PR #292, opened 2026-05-24
feat: Podman + nerdctl runtime support + PR-time arm64 docker smoke

NetVar1337:
## Summary

Two coordinated changes that improve cross-platform compatibility for operators on non-default setups:

1. **Container runtime detection** — the launcher now works with **Podman 4.4+** and **nerdctl** in addition to Docker. Auto-detection with an explicit `DECEPTICON_CONTAINER_RUNTIME` override.
2. **PR-time arm64 Docker build smoke** — catches Raspberry Pi / Apple Silicon / AWS Graviton breakage at PR time, not only at release tag.

### Detection layer — `clients/launcher/internal/runtime/`

New package with `Detect()` returning a populated `Runtime` struct: `Name`, `Bin`, `ComposeArgs`, `DockerHost`, `Rootless`, `Version`.

Selection order (first hit wins):

1. `$DECEPTICON_CONTAINER_RUNTIME` override (`docker` / `podman` / `nerdctl`)
2. `docker` on PATH and `docker info` succeeds (covers real Docker, Lima/Colima/Rancher Desktop shims, Docker Desktop)
3. `podman` on PATH and `podman info` succeeds (Podman 4+, rootless or rootful)
4. `nerdctl` on PATH and `nerdctl info` succeeds (containerd-native, Rancher Desktop containerd engine)

For Podman, the layer additionally:

- Picks `podman compose` (4.4+ built-in) when available; falls back to the `podman-compose` Python wrapper.
- Discovers the API socket (`$XDG_RUNTIME_DIR/podman/podman.sock` for rootless, `/run/user/$UID/podman/podman.sock`, or `/run/podman/podman.sock`).
- Exposes the socket via `DOCKER_HOST` in the compose subprocess env so nested Docker-API consumers (testcontainers, kubectl with docker-shim, GitOps tooling) keep working transparently.
- Detects rootless mode via `podman info --format {{.Host.Security.Rootless}}`.

### Compose refactor — `clients/launcher/internal/compose/compose.go`

Every `exec.Command("docker", ...)` is now `exec.Command(c.Runtime.Bin, ...)`. The `baseArgs()` shape is built from `c.Runtime.ComposeArgs` (`["compose"]` for Docker/Podman-4.4+, `[]` for `podman-compose`).

Backward-compatible: when callers instantiate `Compose{}` directly (as the legacy tests do) without going through `New()`, `baseArgs()` falls back to the Docker-shaped argv. `TestBaseArgs` and `TestAllProfiles` pass unchanged.

### Installer — `scripts/install.sh`

Preflight now accepts Docker OR Podman OR an explicit `DECEPTICON_CONTAINER_RUNTIME` override. The error message lists all three runtimes when none is found.

```bash
# Auto-detect (prefers Docker, falls back to Podman)
curl -fsSL https://decepticon.red/install | bash

# Force Podman even when Docker is also installed
DECEPTICON_CONTAINER_RUNTIME=podman curl -fsSL https://decepticon.red/install | bash
```

### Docs — `docs/setup-guide.md`

New sections **Using Podman instead of Docker** and **Using nerdctl** with rootless caveats (privileged-port handling for c2-sliver, SELinux mount label disable, `host.docker.internal` mapping equivalence).

## PR-time arm64 Docker smoke

The release pipeline (`.github/workflows/release.yml`) already builds every image multi-arch (`linux/amd64,linux/arm64`) with cosign signatures + CycloneDX SBOMs. **But the PR pipeline only builds amd64**, so an arm64 break only surfaces at release tag — way too late.

### New `docker-build-arm64` job in `ci.yml`

Builds `linux/arm64` for the two images that build fastest under QEMU (`cli` + `langgraph` — both clock in around 3-5 min under QEMU). `sandbox` and `web` take 20+ min under QEMU and would dominate PR CI wall time; they continue to be validated via the release pipeline only.

```yaml
- uses: docker/setup-qemu-action@v4
- uses: docker/setup-buildx-action@v4
- uses: docker/build-push-action@v7
  with:
    platforms: linux/arm64
    load: false                      # cross-arch images can't load into amd64 docker; just validate
    cache-from: type=gha,scope=arm64
    cache-to: ...,scope=arm64        # separate cache slot per arch
```

## Tests

`runtime_test.go` (new, 7 cases) covers `Apply()` (DOCKER_HOST injection + user-override respect), `String()` rendering, `Detect()` invalid-override rejection, and `podmanSocket()` empty-when-missing.

All existing compose tests still pass (`TestBaseArgs`, `TestAllProfiles`, `TestImageTag` unchanged behavior; `TestNew` was pre-existing Windows-path-flaky, passes on Linux).

## Risk

- **Docker users**: zero behavioral change. The detection layer picks `docker` first and falls through to the same `docker compose -f ... up` argv it always emitted.
- **Podman users (new)**: needs Podman 4.4+ with `podman compose` (or `podman-compose` wrapper). Rootless Podman has documented caveats for the c2-sliver profile (privileged ports).
- **arm64 CI lane**: adds ~3-5 min to PR runtime for PRs that touch Docker files. Cached aggressively to keep the marginal cost low.

## Relationship to other PRs

- Builds cleanly on `upstream/main`.
- Compose tests: `TestBaseArgs` updated implicitly via the new fallback (no test change needed). The `TestNew` Windows-path failure is pre-existing in main and unrelated to this PR.
- Does not touch any Python runtime code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -349,6 +349,48 @@ jobs:
           path: trivy-${{ matrix.image }}.sarif
           if-no-files-found: ignore
 
+  # Cross-architecture (linux/arm64) smoke build for Raspberry Pi /
+  # Apple Silicon targets. Releases already build multi-arch images, but
+  # we want PR-time signal so an arm64 break doesn't only surface at
+  # release time. Limited to the two images that are smallest under
+  # QEMU (cli + langgraph build in ~3-5 min; sandbox/web take 20+ min
+  # under QEMU and would dominate PR CI wall time — they're validated
+  # via the release pipeline only).
+  #
+  # No Trivy step here: `load: false` means the foreign-arch image stays
+  # in the buildx cache and isn't materialized into the host Docker, so
+  # there's nothing for Trivy to scan. The release pipeline scans the
+  # arm64 images post-push.
+  docker-build-arm64:
+    name: Docker arm64 smoke (${{ matrix.image }})
+    needs: changes
+    if: needs.changes.outputs.docker-images != '[]'
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image: [cli, langgraph]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: docker/setup-qemu-action@v4
+      - uses: docker/setup-buildx-action@v4
+      - name: Build linux/arm64 image (no push)
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: containers/${{ matrix.image }}.Dockerfile
+          push: false
+          # `load: false` with --platform linux/arm64 keeps the image in
+          # the buildx cache without trying to load a foreign-arch image
+          # into the host's local Docker (which would error on amd64
+          # runners). The build itself still validates the entire
+          # Dockerfile path on the target arch.
+          load: false
+          platforms: linux/arm64
+          tags: decepticon-${{ matrix.image }}:ci-arm64
+          cache-from: type=gha,scope=arm64
+          cache-to: type=gha,mode=${{ github.event_name == 'push' && 'max' || 'min' }},scope=arm64
+
   # Security scans always run regardless of which paths changed —
   # vulnerability databases update independently of repo state.
   security:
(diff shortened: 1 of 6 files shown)


PurpleCHOIms:
Superseded by #318 (merged as 4e46164) — re-applied on current main. compose.go RemoveOrphanedCLI conflict with #285 resolved by keeping both the stack-anchored container filter and the runtime abstraction. Closing this original.

PR #293, opened 2026-05-24
ci: pre-commit hook gate (file hygiene + shellcheck + hadolint + typos)

NetVar1337:
## Summary

Adds a curated set of pre-commit hooks plus a CI job that runs them on every PR. The hook selection is tuned to add real value without firing on pre-existing files (no mass reformat required to merge).

### Generic file hygiene (\pre-commit/pre-commit-hooks\)

- **\check-json\** — catches malformed JSON before runtime
- **\check-case-conflict\** — critical for cross-platform compatibility. macOS and Windows default to case-insensitive filesystems, so \Foo.py\ + \oo.py\ silently collide for those contributors. Linux-only contributors might not notice
- **\check-symlinks\** + **\destroyed-symlinks\** — catches broken symlinks at PR time vs runtime
- **\check-executables-have-shebangs\** + **\check-shebang-scripts-are-executable\** — pair that catches the most common file-permission drift

### Shell (\shellcheck-py\)

\shellcheck\ against every shell script with \--severity=warning\. Style warnings stay non-blocking via documented excludes:

- \SC2086\ (word-splitting) — pre-existing throughout, tracked follow-up
- \SC2155\ (declare-and-assign) — pre-existing idiom in \install.sh\
- \SC1091\ (dynamic source) — false positive on optional includes

Real bugs (unset variables, \[[\ vs \[\ semantics on busybox sh, dead code, etc.) still fail the gate.

### Dockerfiles (\hadolint\)

\hadolint\ on \containers/*.Dockerfile\ with sensible ignores:

- \DL3008\ apt-get version pinning — we use base-image repo pinning
- \DL3018\ apk version pinning — Alpine doesn't expose snapshot URLs
- \DL3015\ always \--no-install-recommends\ — many cases legitimately need recommends
- \DL3059\ consecutive \RUN\ — deliberately combined for cache control

Catches missing \USER\, missing \--chown\, etc.

### Typos (\crate-ci/typos\)

Fast Rust-based typo finder. New repo-root **\.typos.toml\** whitelists ~50 offensive-security tool names (impacket, kerberoasting, certipy, mythic, frida, etc.) and skips vendored / generated content (lock files, \_corpus/\, etc.).

### Existing hooks (kept)

- \
uff\ + \
uff-format\ (Python lint + format)
- \gitleaks\ (secrets scan)
- \asedpyright\ (local) — Python type-check
- Plus the generic hygiene set (trailing-whitespace, end-of-file-fixer, check-yaml, check-toml, check-merge-conflict, check-added-large-files, detect-private-key, mixed-line-ending)

## New \pre-commit\ CI job

\.github/workflows/ci.yml\ — runs the hook chain across the whole tree on every PR.

Hooks that need a language toolchain (\asedpyright\, \go-vet\) run in their own jobs already, so the pre-commit gate skips them via \SKIP=basedpyright,go-vet\.

Cached on \.pre-commit-config.yaml\ hash so first run installs, every subsequent run loads from \~/.cache/pre-commit\ (typical 5-10s warm vs 30-45s cold).

## Local pre-push hooks

\asedpyright\ + \go-vet\ are kept as local hooks gated to \stages: [pre-push]\ so contributors without uv / go installed can still commit. They run automatically on push and in CI's matching jobs.

## Risk

Adding hooks can fire on pre-existing issues. The selected hook set was tuned to NOT fire on the current tree:

- shellcheck severity=warning + 3 documented excludes
- hadolint ignores intentional Dockerfile patterns
- typos has the 50-name allowlist for offensive-security jargon
- \markdownlint\ / \yamllint\ / \prettier\ deliberately **NOT** added in this PR — they would fire on many existing files; reserved for a follow-up reformat PR

If the gate fails on a pre-existing pattern I missed, the fix should land in this PR's review cycle (either expand the ignore list or fix the offender — case-by-case decision).


PurpleCHOIms:
Superseded by #322 (merged as 4bfd1eb) — re-applied on current main with a 2-commit structure: (1) renormalize 45 CRLF files to LF (pre-existing repo debt #293 surfaced), (2) the hook gate + tuning fixes for the tree as it stands today (extend-identifiers for NDA/AKS/RTO, ICS-OT exclude, mixed-line-ending exclude for .ps1/.bat, drop unused install.sh DECEPTICON_RUNTIME). All hooks pass, no silent auto-fix. Closing this original.

PR #294, opened 2026-05-24
fix: surface 4 silent exception swallows + cap unbounded httpx timeouts (CVE lookups)

NetVar1337:
## Summary

Bug-hunter sweep (\
uff -S110\ + manual review) found 4 places where exceptions were silently swallowed. Each silently degraded observability in a way that makes on-call debugging hard. Also caps three unbounded \httpx\ timeouts in CVE-lookup paths that were causing intermittent false-negatives against the NVD API.

## Silent-swallow fixes (4)

| File | Symptom that was hidden | Fix |
|---|---|---|
| \gents/prompts/__init__.py:455\ | Claude-4 compatibility shim failure → prompt ships without corrections; later behavioral drift has no trace | Log at debug with agent name + traceback |
| \	ools/research/_state.py:close_store()\ | Neo4j driver close-failure at process teardown swallowed; teardown-order bugs invisible | Log at debug |
| \	ools/research/chain.py:compute_path_score()\ | APOC procedure unavailable → silent fallback to plain \shortestPath\; operators don't know the cheaper plan ran | Log at debug |
| \	ools/research/cve.py:_Cache.__init__()\ | \texit.register()\ failure → cache silently loses dirty state on exit | Log at debug |

Each uses \log.debug\ (not warning/error) because every site is a documented fallback path, not a hard failure. Production logs stay quiet; \DECEPTICON_LOG_LEVEL=DEBUG\ reveals which fallbacks fired.

## \httpx\ timeout caps (3)

\cve.py\ had three \httpx.AsyncClient()\ instantiations with no explicit timeout. httpx defaults to 5s per stage, which the public NVD API regularly exceeds (rate-limit windows, regional CDN slow paths). False-negatives on legitimate CVE lookups dropped to zero in local testing after the cap was raised:

\\\
lookup_cve():       30s end-to-end / 10s connect (NVD + EPSS)
lookup_cves():      30s end-to-end / 10s connect (batch NVD)
lookup_package():   15s end-to-end /  5s connect (OSV is reliably fast)
\\\

\httpx.Timeout(N.0, connect=M.0)\ gives a generous read budget while keeping connect tight (a dead endpoint should fail fast).

## Verified

- \
uff check --select S110 packages/decepticon/decepticon/\: was 4 errors, now **0**
- \
uff check\ on changed files: clean
- \
uff format --check\ on changed files: clean
- No behavior change for the happy path

## Why these (and not the other 19 \-S\ findings)

The full ruff -S sweep returned 19 findings; the other 15 are false positives in this codebase:

- \S324\ (md5/sha1 weak hash): every site uses the hash for cache keys / short IDs (\[:6]\, \[:8]\, \[:12]\, \[:16]\), not security
- \S603\ (subprocess untrusted-input): every call uses list-form argv (no shell injection) and operates on trusted internal commands
- \S101\ (assert): used for invariants in non-test code; legitimate
- \S106\ (hardcoded password): \secret_type="krb5asrep"\ is a labeling string, not a credential

Those 4 categories could individually justify \# noqa\ annotations to keep CI quiet, but that's separate operator-experience polish, not a bug fix. Tracked for a follow-up.

## Risk

Minimal. Four added log calls; three timeout values that increase from 5s to 15-30s. Happy-path behavior unchanged.

### DIFF
--- a/packages/decepticon/decepticon/agents/prompts/__init__.py
+++ b/packages/decepticon/decepticon/agents/prompts/__init__.py
@@ -30,10 +30,13 @@ Usage:
 
 from __future__ import annotations
 
+import logging
 import os
 from datetime import datetime, timezone
 from functools import lru_cache
 from pathlib import Path
+
+log = logging.getLogger("decepticon.agents.prompts")
 
 _PROMPTS_DIR = Path(__file__).parent
 
@@ -458,7 +461,14 @@ def load_prompt(
         prompt = apply_compat_for_role(prompt, name)
     except Exception:
         # Fail soft: never break prompt loading because of the compat shim.
-        pass
+        # A failure here is recoverable — the prompt simply ships without
+        # the Claude-4 compat tweaks. Log at debug so the divergent prompt
+        # is traceable when diagnosing model-specific behavioral drift.
+        log.debug(
+            "prompts.load_prompt(%r): claude4_compat shim failed; shipping uncorrected prompt",
+            name,
+            exc_info=True,
+        )
 
     # Plugin + explicit prompt overrides apply last so prepend/append wrap
     # the fully-assembled prompt (language policy + compat shim included).
--- a/packages/decepticon/decepticon/tools/research/_state.py
+++ b/packages/decepticon/decepticon/tools/research/_state.py
@@ -39,7 +39,11 @@ def close_store() -> None:
         try:
             _store.close()
         except Exception:
-            pass  # Neo4j not configured — no-op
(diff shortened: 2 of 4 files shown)


PurpleCHOIms:
Superseded by #323 (merged as 91643da) — scope-trimmed to only the cve.py changes (atexit silent-swallow + 3 httpx timeout caps). The other 3 silent-swallows in this PR (_state.py close_store, chain.py compute_path_score, prompts Claude-4 shim) were already applied to main with identical intent via #315/#316. Closing this original.

PR #296, opened 2026-05-25
fix(tools/bash): inherit proxy env vars into sandbox tmux sessions

NetVar1337:
## Summary

Engagement traffic that should flow through an intercepting proxy (Caido / Burp / ZAP) was bypassing it because `HTTPS_PROXY` and friends set in the orchestrator process never reached the tmux shells where `bash` commands actually run. This PR adds a strict allowlist of parent env vars that `TmuxSessionManager` syncs into the live tmux shell during `initialize()`.

## Behavior change

When `initialize()` runs (either reusing a cached pane or creating a fresh session), after the PS1 marker is installed the manager sends a single `export` line containing only:

- Proxy variables: `HTTPS_PROXY`, `HTTP_PROXY`, `NO_PROXY`, `ALL_PROXY` (and lowercase variants)
- Any variable whose name starts with `DECEPTICON_`

No other parent environment is forwarded. Values are quoted with `shlex.quote`; variable names are re-validated against a POSIX-shell-name regex before emission as defense-in-depth. Sync errors are logged at WARNING and swallowed so sandbox execution survives transient tmux failures.

## Files

| File | Change |
|---|---|
| `packages/decepticon/decepticon/sandbox_kernel/tmux.py` | Add `_PROXY_ENV_KEYS` / `_DECEPTICON_ENV_PREFIX` / `_VALID_ENV_NAME` constants, `_allowed_passthrough_env()` and `_shell_export_command()` helpers, `TmuxSessionManager._sync_passthrough_env()` method, and wire it into `initialize()` on both cached-alive and fresh-session paths. |
| `packages/decepticon/tests/unit/tools/bash/test_proxy_env.py` | 15 unit tests covering the allowlist semantics, shell-quoting (incl. injection-attempt round-trips), no-op-on-empty, error swallowing, and both `initialize()` wiring paths. Lowercase-proxy variant skipped on Windows where env names are case-insensitive. |

## Non-goals

- No change to public `bash` / `bash_output` / `bash_kill` tool surface
- No arbitrary parent environment forwarded
- No new dependencies

## Verification

- `uv run pytest packages/decepticon/tests/unit/tools/bash/test_proxy_env.py -q` → 14 passed, 1 skipped
- `uv run pytest packages/decepticon/tests/unit/tools/bash -q` → 34 passed, 1 skipped (no regressions in existing bash module)
- `uv run ruff check packages/decepticon/decepticon/sandbox_kernel/tmux.py packages/decepticon/tests/unit/tools/bash/test_proxy_env.py` → All checks passed
- LSP diagnostics on changed files: clean


NetVar1337:
Superseded by #350 — the consolidated mega-PR that bundles this change together with the rest of the open backlog into one branch. #350 is green on all five required status checks and is mergeable (`CLEAN`); merging the set as a single `--no-ff` commit preserves each PR's authorship in the merge history.

Closing in favor of #350. No branch was deleted — this can be reopened if the consolidation ever needs to be split back out.

PR #303, opened 2026-05-25
feat(runtime): append-only engagement events.jsonl event log

NetVar1337:
## Summary

Add `decepticon.runtime.event_log` so the orchestrator (and a future `resume <engagement_id>` command) can durably record what happens during a run. The log survives crashes / Ctrl-C and is structured as a line-oriented append-only file under `engagements/<id>/events.jsonl`.

## Public surface

| Symbol | Purpose |
|---|---|
| `EventType` (str-enum) | engagement.start / .end / .checkpoint, agent.turn, tool.call / .result, llm.call / .response, finding.created, opplan.update |
| `EngagementEvent` (frozen dataclass) | ts / type / agent / payload + JSON round-trip |
| `EventLog(workspace_root, engagement_id)` | per-engagement writer; safe under concurrent threads + processes |
| `read_events(path)` | yields events in order; silently skips malformed / torn lines |

## Atomicity / durability

- `append()` is safe under concurrent threads (in-process `threading.RLock`) **and** concurrent processes (cross-platform file lock: `fcntl.flock` on POSIX, `msvcrt.locking` on Windows).
- Each write is followed by `flush()` + `os.fsync()` before the lock is released.
- A torn write by a killed process cannot corrupt earlier events — the lock blocks any second writer until the current line is fully flushed.
- `read_events()` silently skips malformed lines so a process killed mid-write cannot make the whole log unreadable.
- Event types are stored as `str` on disk so a newer Decepticon can add a type without breaking older readers.

## Deliberately deferred (follow-up PRs)

- Wiring `EventLog` into `decepticon.middleware.opplan` / `agents.standard.decepticon` so emit sites actually call `append()` at the natural boundaries.
- The `--resume <id>` CLI plumbing across the Go launcher (`clients/launcher/`), TS REPL (`clients/cli/`), and the Python middleware. Multi-layer wiring is its own PR.
- `engagement.checkpoint` emission from the graceful-shutdown handler (already lands in a separate PR; it appends to events.jsonl iff the file exists, so dependency ordering is forward-only).

The Python library + tests land first as the clean, dependency-free ground truth the integration PRs can adopt.

## Files

| File | Δ |
|---|---|
| `packages/decepticon/decepticon/runtime/__init__.py` | exports public API |
| `packages/decepticon/decepticon/runtime/event_log.py` | new — 229 lines |
| `packages/decepticon/tests/unit/runtime/__init__.py` | empty package marker |
| `packages/decepticon/tests/unit/runtime/test_event_log.py` | new — 137 lines, 11 tests |

## Verification

- `uv run pytest packages/decepticon/tests/unit/runtime/test_event_log.py -q` → **11 passed**
- `uv run ruff check packages/decepticon/decepticon/runtime packages/decepticon/tests/unit/runtime` → All checks passed
- LSP diagnostics on changed files: clean

## Test coverage

One-line-per-event JSONL output, in-order read, `append()` return value, `str`-typed event names accepted, malformed-line skipping during read, missing-file handling, `EngagementEvent` JSON round-trip, invalid-JSON returns `None`, per-engagement path layout, two `EventLog` instances appending to the same file in sequence, and a **200-thread concurrent atomicity smoke test** that asserts every line parses and no events were lost or interleaved.

### DIFF
--- a/packages/decepticon/decepticon/runtime/__init__.py
+++ b/packages/decepticon/decepticon/runtime/__init__.py
@@ -1,19 +1,35 @@
 """Runtime infrastructure for the Decepticon orchestrator process.
 
-Currently exposes the append-only engagement event log; future
-primitives (graceful shutdown, resume) will live alongside it.
+Each runtime module (event log, graceful shutdown, future resume, ...)
+ships in its own PR. This package init imports every known module
+best-effort so each PR can land in any order without colliding on
+this file.
 """
 
-from decepticon.runtime.event_log import (
-    EngagementEvent,
-    EventLog,
-    EventType,
-    read_events,
-)
+_exports: list[str] = []
 
-__all__ = [
-    "EngagementEvent",
-    "EventLog",
-    "EventType",
-    "read_events",
-]
+try:
+    from decepticon.runtime.event_log import (  # noqa: F401  # re-exported via __all__
+        EngagementEvent,
+        EventLog,
+        EventType,
+        read_events,
+    )
+
+    _exports += ["EngagementEvent", "EventLog", "EventType", "read_events"]
+except ImportError:
+    # event log module not present — another runtime PR may be mid-merge
+    pass
+
+try:
+    from decepticon.runtime.shutdown import (  # noqa: F401  # re-exported via __all__
+        LangGraphState,
+        install_shutdown_handlers,
+    )
+
+    _exports += ["LangGraphState", "install_shutdown_handlers"]
+except ImportError:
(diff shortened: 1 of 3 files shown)


NetVar1337:
Superseded by #350 — the consolidated mega-PR that bundles this change together with the rest of the open backlog into one branch. #350 is green on all five required status checks and is mergeable (`CLEAN`); merging the set as a single `--no-ff` commit preserves each PR's authorship in the merge history.

Closing in favor of #350. No branch was deleted — this can be reopened if the consolidation ever needs to be split back out.

PR #319, opened 2026-05-26
chore: renormalize line endings to match .gitattributes

NetVar1337:
## Problem

The repo's .gitattributes declares LF for `*.md`, `*.json`, `*.ts`, `*.tsx`, `*.py`, `*.yml`, Dockerfiles, etc., but **45 tracked blobs are stored with CRLF**. The mismatch causes:

- `git status` shows these 45 files as permanently modified on every fresh clone (on hosts without `core.autocrlf=true` masking it)
- `git diff` produces a misleading 8000-line "every line changed" diff for anyone who touches one of these files
- New contributors can't tell their real changes from the line-ending noise
- Pre-commit hooks/CI that hash file content may flap

Repro on a clean clone:
```bash
git clone https://github.com/PurpleAILAB/Decepticon.git
cd Decepticon
git config core.autocrlf false   # or run on Linux/macOS
git status                       # 25+ files marked modified, untouched
```

## Fix

Single `git add --renormalize .` commit. Pure whitespace change:

- **45 files changed**
- **8026 insertions, 8026 deletions** (1:1, every changed line is the same line with CRLF → LF)
- `package-lock.json` shrinks by exactly 14,044 bytes (the stripped CR characters)
- No source content modified

After this lands, fresh clones will be clean.

## Verification

```bash
git diff -w HEAD~1 HEAD   # produces only the package-lock.json binary delta
                           # (it has -diff attribute, so git treats it binary;
                           #  the 14044-byte size drop is the CR strip)
```


PurpleCHOIms:
Already applied via #322 (the LF renormalization commit 8ed608c on revival/293 → merged as 4bfd1eb). The numbers match exactly — 45 files, 8026 insertions / 8026 deletions, package-lock.json shrinks 14044 bytes. Fresh clones are already clean. This PR (based on e1afba6) now shows a -732-line net diff because it would regress the 5 PRs merged since (#322/#323/#325/#326/#327). Closing as duplicate.

PR #324, opened 2026-05-26
fix(streaming): preserve StreamingRunnable across deepagents .with_config()

PurpleCHOIms:
## Summary

Subclass `RunnableBinding` so `StreamingRunnable` survives `deepagents.middleware.subagents._get_subagents()`'s `compiled["runnable"].with_config({...})` call. Without this, the wrapper is silently dropped before dispatch and **zero `subagent_*` custom events** reach the LangGraph Platform HTTP stream.

## The bug

`StreamingRunnable` was a plain class with `__getattr__` forwarding to the wrapped graph. deepagents normalises each spec at registration time with:

```python
runnable = compiled["runnable"].with_config({"metadata": {"lc_agent_name": ...}, "run_name": ...})
```

`with_config` is not defined on the wrapper, so `__getattr__` forwards the call to the inner compiled graph. The returned `RunnableBinding` has `bound = inner_raw_graph`, not the streaming wrapper — every subsequent `await subagent.ainvoke(...)` goes straight to the unwrapped binding.

The CLI/UIRenderer channel still worked because that path is set via a contextvar before `.with_config(...)` ever runs (and reads from the same contextvar inside the subagent's nodes/tools). The LangGraph Platform HTTP `stream_mode=["custom"]` path was always broken; no `subagent_*` event ever reached it.

## The fix

Subclass `RunnableBinding`. Its `with_config()` reconstructs via `self.__class__(...)`, so the subclass identity is preserved across deepagents' registration step.

- `bound` / `name` / `config` now live on the Pydantic fields `RunnableBinding` already exposes.
- `_runnable` and `_name` become read-only properties for back-compat with existing callers and the regression tests.
- `__getattr__` is removed — `RunnableBinding` covers the attributes deepagents and downstream middleware actually touch.
- The constructor accepts both the legacy positional `StreamingRunnable(runnable, name)` shape used at registration sites and the keyword-only `bound=..., config=..., kwargs=...` shape `RunnableBinding.with_config()` uses internally.

## Verification

Against `langgraph` 1.2.1 / `langchain-core` 1.4.0 / `deepagents` 0.6.3 on a LangGraph Platform self-hosted run:

| | Before patch | After patch |
|---|---|---|
| `custom` SSE frames for one `task("recon", …)` delegation | **0** | **5** (`subagent_start`, `subagent_tool_call` for `bash`, `subagent_tool_result`, `subagent_message`, `subagent_end`) |
| `StreamingRunnable.ainvoke()` invoked | no | yes |
| CLI UIRenderer channel | works | works (unchanged) |

The seven regression tests in `tests/unit/core/test_subagent_streaming.py` pass. The `CountingRunnable` shim was a duck-typed class; `RunnableBinding.bound` now validates `isinstance(Runnable)`, so the fake is updated to subclass `Runnable`. This actually brings the test closer to production — deepagents only ever hands real compiled graphs through this path.

## Test plan

- [x] `pytest packages/decepticon/tests/unit/core/test_subagent_streaming.py` — all 7 pass
- [x] End-to-end: send a fresh kickoff to a Decepticon supervisor on a LangGraph Platform self-hosted instance and capture `runs/stream?stream_mode=["custom"],stream_subgraphs=true`. Five `subagent_*` custom frames now appear inside the parent stream (root `event: custom`).
- [x] CLI/UIRenderer path unchanged (renderer contextvar still receives `on_subagent_*` callbacks during invoke).

## References

- `langchain-core` `runnables/base.py` — `RunnableBinding.with_config` reconstructs via `self.__class__(...)`.
- `deepagents` `middleware/subagents.py` `_get_subagents` line 97 — the `.with_config(...)` call this patch is reactive to.

### DIFF
--- a/packages/decepticon/decepticon/core/subagent_streaming.py
+++ b/packages/decepticon/decepticon/core/subagent_streaming.py
@@ -16,10 +16,15 @@ Architecture:
   → emits events via both channels
   → returns same result as invoke() for SubAgentMiddleware compatibility
 
-IMPORTANT: LangGraph runs the agent graph asynchronously (all middleware uses
-awrap_* methods). SubAgentMiddleware's atask() calls subagent.ainvoke(), so
-the ainvoke() method MUST be implemented here — otherwise __getattr__ delegates
-to the underlying runnable's ainvoke(), bypassing all streaming logic.
+Why a RunnableBinding subclass:
+  deepagents.middleware.subagents._get_subagents() normalises every spec by
+  calling `compiled["runnable"].with_config({"metadata":..., "run_name":...})`
+  and stores the *returned* object as the subagent. If StreamingRunnable were
+  a plain class with __getattr__ forwarding, with_config would fall through
+  to the inner compiled graph and return a RunnableBinding wrapping the bare
+  graph — silently dropping this wrapper before dispatch. RunnableBinding's
+  built-in with_config() reconstructs `self.__class__(...)`, so subclassing
+  preserves the wrapper across deepagents' registration step.
 """
 
 from __future__ import annotations
@@ -32,6 +37,7 @@ from typing import Any, Callable
 
 from langchain_core.messages import AIMessage
 from langchain_core.messages.tool import ToolCall
+from langchain_core.runnables import Runnable, RunnableBinding
 
 log = logging.getLogger("decepticon.subagent_streaming")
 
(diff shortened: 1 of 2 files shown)


PurpleCHOIms:
Superseded by #327 (merged as 38432a2) — re-applied on current main via cherry-pick of the real fix commit e9f805b (the original PR head was a merge-commit, so the standard cherry-pick path failed). Clean apply, 7 regression tests pass, pre-commit gate green. Closing this original.

PR #336, opened 2026-05-26
fix(sandbox): reap zombie children on SIGCHLD and clean up tmux sessions

pol2up:
## Problem

`decepticon-sandbox` accumulates `<defunct>` `tmux` + `bash` processes during long-running engagements (observed: 4 zombies in one session). The HTTP daemon spawns many short-lived subprocesses (tmux subcommands, `sh -c` for `execute()`), and while most are reaped by their immediate `subprocess.run(...)` caller, the tmux server itself, the bash shells inside `tmux new-session`, and any shell-spawned grandchildren are not waited on by the daemon. When the tmux server exits, its children get reparented to the daemon (PID 1 of the sandbox container) and the daemon never calls `waitpid` on them, so they linger as zombies for the daemon's lifetime. Under sustained use the per-process PID table eventually fills and `fork()` starts failing with `EAGAIN`.

### Part 1 — SIGCHLD reaper

New `decepticon.sandbox_server.reaper` module installs a signal handler that loops `os.waitpid(-1, os.WNOHANG)` until no more zombies remain. Wired from the FastAPI `lifespan` startup hook **before** the backend warms up (which is the first thing that spawns subprocesses). The reaper is also registered on the asyncio event loop via `loop.add_signal_handler()` as a belt-and-braces drain that runs in the event-loop thread.

The reaper is reentrant-safe — `os.waitpid` with `WNOHANG` in a loop is the standard async-signal-safe Unix idiom and we do no allocation in the signal frame beyond what CPython's signal machinery already does.

### Part 2 — tmux session cleanup on shutdown

New `SandboxBase.kill_all_sessions()` iterates every `TmuxSessionManager` the sandbox has handed out and runs `tmux kill-server` on each; the lifespan shutdown hook invokes it so no tmux server outlives the daemon. `kill-server` (rather than `kill-session`) is the right hammer for shutdown: it guarantees the tmux process **and** every bash child it spawned exit, leaving nothing for the kernel to reparent.

Both code paths are best-effort with errors logged not raised — one stuck session must not block the rest from being torn down.

## Tests

- **`tests/unit/sandbox_server/test_reaper.py`** (5 tests): reaper in isolation, including a real `fork() + parent exits + grandchild exits` scenario to verify the SIGCHLD handler drains reparented grandchildren.
- **`tests/unit/backends/test_session_log.py`** (3 new tests): `kill_all_sessions` contract — every manager killed, failure-isolation, no-op when empty.
- **`tests/unit/sandbox_server/test_app.py`** (4 new tests): lifespan wiring — startup installs the reaper, shutdown calls `kill_all_sessions` + `reap_zombies`, shutdown swallows tear-down errors.

```
$ uv run pytest packages/decepticon/tests/unit/sandbox_server/ packages/decepticon/tests/unit/backends/ packages/decepticon/tests/unit/sandbox_kernel/
116 passed
```

## Risk

Low — sandbox-internal process management. The reaper only mutates global signal state (recoverable on shutdown) and the kill-server-on-shutdown path only fires during daemon teardown when the sandbox is going away anyway. No production caller behaviour changes.

### DIFF
--- a/packages/decepticon/decepticon/sandbox_kernel/base.py
+++ b/packages/decepticon/decepticon/sandbox_kernel/base.py
@@ -356,5 +356,48 @@ class SandboxBase(BaseSandbox):
             self.reset_session_log_offset(session, effective_workspace)
             self._jobs.remove(session, key=manager_key)
 
+    def kill_all_sessions(self) -> int:
+        """Kill every tmux session this sandbox has handed out a manager for.
+
+        Returns the number of sessions kill-session was invoked on. Used by
+        the sandbox daemon's shutdown hook to make sure tmux servers don't
+        outlive the daemon process and leave behind zombie ``bash``/``tmux``
+        children when their parent disappears.
+
+        Best-effort: every kill is wrapped in try/except so one stuck
+        session doesn't block the rest from being torn down.
+        """
+        with self._managers_lock:
+            managers = list(self._managers.values())
+            keys = list(self._managers.keys())
+            self._managers.clear()
+        killed = 0
+        for mgr in managers:
+            try:
+                # Each TmuxSessionManager runs its own ``tmux -L <name>``
+                # socket (see TmuxSessionManager._tmux), so kill-server
+                # terminates exactly the tmux process backing this session
+                # — including reaping the bash child the new-session
+                # command spawned. This is more thorough than kill-session
+                # for the shutdown path: even if the session was already
+                # detached, the tmux server itself exits.
+                mgr._tmux(["kill-server"], timeout=5)
+                killed += 1
+            except (RuntimeError, subprocess.TimeoutExpired, OSError) as e:
+                log.debug(
+                    "kill-server failed for session '%s' during shutdown: %s",
+                    _safe_log(mgr.session),
+                    _safe_log(e),
+                )
+        # Reset all per-session tracking state alongside the manager
+        # cache so a subsequent restart starts from a clean slate.
+        with TmuxSessionManager._init_lock:
+            for mgr in managers:
+                TmuxSessionManager._initialized.discard(mgr.session)
+        for key in keys:
+            with self._log_offsets_lock:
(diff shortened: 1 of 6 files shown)


PurpleCHOIms:
Post-merge review found a **critical correctness regression** introduced by the SIGCHLD reaper in this PR.

The daemon runs as PID 1 (the sandbox entrypoint `exec`s it), so reparented tmux/bash grandchildren genuinely do zombify when their tmux server exits — that problem is real. But the fix installs a process-wide `signal.signal(SIGCHLD, reap_zombies)` (plus an asyncio `add_signal_handler`) whose `os.waitpid(-1, WNOHANG)` loop **races with the daemon's own `subprocess.run` calls** (`sandbox_kernel/base.py::execute()`, `tmux.py::_run()`).

When the reaper reaps a child before `subprocess.run`'s own `Popen.wait()`, `os.waitpid(self.pid)` raises `ChildProcessError`; CPython's `Popen._try_wait` swallows it and assumes exit status 0 — **real non-zero exit codes are silently clobbered to success.**

Measured with the exact code from this PR:
- single-threaded: **200/200** commands returned a bogus exit code
- under the FastAPI worker-thread model (sync endpoints run in anyio threads): **~30%** corruption

Impact: `execute()` returns the wrong `exit_code`, and `tmux.py::_run()`'s `if result.returncode != 0: raise TmuxCommandError` never fires — real tmux failures are masked as success. For an agent that branches on command success, this corrupts decisions.

The idiomatic fix is to let an init process reap orphans instead of a process-wide handler: `init: true` on the sandbox compose service makes tini PID 1, which reaps reparented orphans while `subprocess.run` keeps reaping its own children (exit codes intact). Verified: with `docker --init` the orphan grandchild zombie goes 1 → 0; removing the in-process reaper restores exit codes 200/200 → 0/200.

Fixed in #340, which keeps the race-free `kill_all_sessions()` shutdown cleanup from this PR.

PR #389, opened 2026-05-30
test(backends): cover HTTPSandbox transport — 28%→full

NetVar1337:
Adds 34 unit tests for `backends/http_sandbox.py` (the HTTP-transport sandbox backend), previously the lowest-covered non-POSIX module (28.8%).

Uses `httpx.MockTransport` to drive the full request-build / response-parse path without a live daemon:

- **Retry/back-off** (`_retry_on_connection_error`): immediate success, recovery after transient `ConnectError` with exponential delays, exhaustion re-raises `last_exc`, non-connection errors not swallowed.
- **Client lifecycle**: trailing-slash stripping, lazy + cached `httpx.Client`, bearer-token header vs none, idempotent `close()`, per-instance job tracker.
- **`_request` error path**: HTTP-status → `SandboxError` with 200-char body truncation; retry-then-success.
- **execute / upload / download**: response parsing + defaults, request-timeout margin (+10s), base64 round-trip both directions, per-file error surfacing.
- **tmux surface**: sync + async (`execute_tmux_async` thread delegation), timeout margin.
- **background jobs**: `start_background` local-mirror registration, `poll_completion` none/done/running, missing-mirror re-register, `exit_code` `None`→`-1` fallback.

Test-only change; no source modified. All gates green locally (ruff, format, basedpyright 0 errors, pytest 34 passed).

### DIFF
--- a/packages/decepticon/tests/unit/backends/test_http_sandbox.py
+++ b/packages/decepticon/tests/unit/backends/test_http_sandbox.py
@@ -1,0 +1,424 @@
+"""Unit tests for the HTTP-transport sandbox backend (``HTTPSandbox``).
+
+Every method on ``HTTPSandbox`` forwards a JSON request to the in-container
+``decepticon.sandbox_server`` daemon and parses the JSON reply. The tests
+inject an ``httpx.MockTransport`` so the full request-build / response-parse
+path runs without a live daemon, and exercise the retry/back-off helper and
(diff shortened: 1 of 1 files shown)


NetVar1337:
Closing as a duplicate of #418, which provides equivalent HTTPSandbox transport coverage (same test module path). Keeping #418. Merge #418 instead.

PR #391, opened 2026-05-30
fix(tools/cloud): honor IAM NotAction semantics in policy analyzer

NetVar1337:
## Defect

`analyze_iam_policy` in `packages/decepticon/decepticon/tools/cloud/aws.py` (line 137) aliased `NotAction` to `Action` via `stmt.get("Action") or stmt.get("NotAction")`. AWS `NotAction` has inverted semantics: `{"Effect":"Allow","NotAction":"s3:GetObject","Resource":"*"}` grants **every** action except `s3:GetObject` — effective admin access. The old code iterated only the excluded action, ran it through the primitive matcher, reported nothing, and returned zero findings — a false negative on a common privesc pattern. Conversely, a `NotAction: "iam:CreateAccessKey"` statement would trigger a false-positive CreateAccessKey finding even though that action is the one being withheld.

## Impact

The `iam_policy_audit` tool silently produces wrong results for any policy using `NotAction`. A `NotAction`+`Resource=*` grant (a known IAM privilege-escalation pattern used in real attacker policies) goes entirely unreported.

## Fix

`NotAction` statements are now handled in a dedicated branch before the `Action` loop:
- `Allow` + `NotAction` + `Resource=*` → **critical** finding (near-admin grant)
- `Allow` + `NotAction` + scoped resource → **high** finding (inverted action set on those resources)
- The per-primitive privesc matcher only runs against genuine `Action` lists

## Regression test

`packages/decepticon/tests/unit/cloud/test_aws_notaction.py` — 6 cases covering:
- `NotAction` + `Resource=*` → critical
- `NotAction` + scoped resource → high
- No false-positive primitive match from `NotAction` values
- List-valued `NotAction`
- Regular `Action` path still produces correct findings
- `Deny`+`NotAction` ignored

All 29 tests (6 new + 23 existing cloud tests) green.

No interaction with in-flight PRs (no open PR touches `aws.py` or the cloud test suite).

### DIFF
--- a/packages/decepticon/decepticon/tools/cloud/aws.py
+++ b/packages/decepticon/decepticon/tools/cloud/aws.py
@@ -134,7 +134,47 @@ def analyze_iam_policy(policy: str | dict[str, Any]) -> list[IAMFinding]:
         effect = (stmt.get("Effect") or "Allow").lower()
         if effect != "allow":
             continue
-        for action in _as_list(stmt.get("Action") or stmt.get("NotAction") or "*"):
+
+        not_action_list = stmt.get("NotAction")
+        if not_action_list is not None:
+            excluded = [str(a).lower() for a in _as_list(not_action_list)]
+            resources = _as_list(stmt.get("Resource") or "*")
+            has_star_resource = any(str(r) == "*" for r in resources)
+            resource_summary = ", ".join(str(r) for r in resources)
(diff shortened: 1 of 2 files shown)


NetVar1337:
Closing as a duplicate of #410, which fixes the same IAM NotAction-as-Action inversion and additionally handles the empty-Action case (branches on key presence). Merge #410 instead.

PR #392, opened 2026-05-30
fix(roe): close two target-extraction scope bypasses (URL userinfo + TLD/extension collision)

NetVar1337:
## Summary
The RoE evaluator refuses out-of-scope hosts only for targets returned by `extract_targets()` (`middleware/roe.py:205`). Two extraction gaps let a host evade scope enforcement entirely.

### 1. URL userinfo bypass (security)
`_URL_RE` captured `[^\s/:]+` immediately after `://`, so `curl https://user:pass@evil.example/` captured `user` (dropped as invalid) and the **real host `evil.example` was never evaluated** by RoE. The regex now skips an optional `userinfo@` segment and captures the host, excluding `@`.

### 2. TLD / file-extension collision (security)
`_NON_TARGET_EXTENSIONS` dropped any token whose last label matched a file extension — but `.sh .md .py .pl .pub .zip` are **delegated DNS TLDs**, so `curl https://evil.zip` was silently dropped from scope enforcement. Those real-TLD entries are removed (file-only extensions retained); the stale comment claiming no TLD collides is corrected with a security rationale to prevent regression.

Both fixes err toward over-extraction, matching this module's documented *"false positives are safer than false negatives"* design — a spurious target is operator-overridable; a dropped real one is a silent scope breach.

## Tests
New `tests/unit/middleware/test_command_targets_scope_bypass.py`: userinfo stripping (with port handling), TLD-colliding hosts now extracted, and confirmation that genuine local-file arguments (`key.pem`, `scan.txt`, …) stay excluded. Full existing RoE suite passes (64 tests green locally).

### DIFF
--- a/packages/decepticon/decepticon/middleware/_command_targets.py
+++ b/packages/decepticon/decepticon/middleware/_command_targets.py
@@ -33,7 +33,10 @@ from collections.abc import Callable
 
 _IP_RE = re.compile(r"\b(?:\d{1,3}\.){3}\d{1,3}\b")
 _CIDR_RE = re.compile(r"\b(?:\d{1,3}\.){3}\d{1,3}/\d{1,2}\b")
-_URL_RE = re.compile(r"\b(?:https?|ftp|file|smb|nfs|ssh|rdp|ldaps?)://([^\s/:]+)", re.IGNORECASE)
+_URL_RE = re.compile(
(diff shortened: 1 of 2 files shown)


NetVar1337:
Closing as a duplicate. The two RoE scope-extraction fixes here (URL userinfo + real-TLD/extension collision) are now consolidated into #405, which uses a more robust RFC-3986 urlsplit-based host extraction and additionally normalizes packed-integer/hex IPv4 IMDS encodings. The TLD-collision fix + regression tests from this PR have been folded into #405. Merge #405 instead.

PR #395, opened 2026-05-30
fix(tools/reversing): correct Ghidra recon script entry-point handling

NetVar1337:
## Defect

`_GHIDRA_RECON` (in `scripts.py` line 35) contained:

```python
addrs = [str(a) for a in f.getEntryPoint()]
```

`Function.getEntryPoint()` returns a single `ghidra.program.model.address.Address` object, which is not iterable. In Jython this raises `TypeError: 'GenericAddress' object is not iterable` the first time a real (non-thunk, non-external) function is visited, aborting the entire generated recon script before it can print any function list or imports.

The `addrs` variable was also never referenced anywhere else, making the line purely dead-then-crash code.

## Impact

The `bin_ghidra_script` `@tool` emits this script body for the agent to write to disk and execute via `analyzeHeadless`. Any RE workflow invoking the standard Ghidra recon script produces a runtime crash with no function output.

## Fix

Deleted the broken unused line. `getEntryPoint()` is already printed correctly on the immediately following `print(...)` line — no functional change needed, just removal.

## Regression Test

`packages/decepticon/tests/unit/reversing/test_ghidra_script_entrypoint.py` — 4 new assertions:
- `test_no_iterable_getentrypoint`: asserts the broken `for a in f.getEntryPoint()` pattern is absent (FAIL without fix, PASS with fix)
- `test_addrs_variable_absent`: asserts `addrs` is not in the generated script (FAIL without fix, PASS with fix)
- `test_entrypoint_printed_once`: asserts `getEntryPoint()` appears exactly once — the correct print call (FAIL without fix due to duplicate, PASS with fix)
- `test_binary_path_substituted`: sanity-checks binary path substitution works

### DIFF
--- a/packages/decepticon/decepticon/tools/reversing/scripts.py
+++ b/packages/decepticon/decepticon/tools/reversing/scripts.py
@@ -32,7 +32,6 @@ print("[+] Functions: {{}}".format(fm.getFunctionCount()))
 for f in fm.getFunctions(True):
     name = f.getName()
     if not f.isThunk() and f.isExternal() is False:
-        addrs = [str(a) for a in f.getEntryPoint()]
         print("fn {{}} @ {{}}".format(name, f.getEntryPoint()))
 
 # Dump external imports (likely interesting APIs)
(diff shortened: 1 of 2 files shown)


NetVar1337:
Closing as a duplicate. The Ghidra recon-script getEntryPoint crash fix has been folded into #422 (which also tolerates extra Ghidra fields and makes the script templates brace-safe). Merge #422 instead.

PR #406, opened 2026-05-30
fix(tools/defense): quote values in Splunk SPL converter

NetVar1337:
## Defect

`_field_clause` in `splunk.py` interpolated raw string values (not `quoted`) into wildcard SPL terms for `contains`, `startswith`, and `endswith` modifiers. A value like `Invoke-Mimikatz -DumpCreds` produced `CommandLine=*Invoke-Mimikatz -DumpCreds*`, which Splunk parses as two separate terms (`CommandLine=*Invoke-Mimikatz` AND bare `-DumpCreds*`), silently changing the detection logic. Values containing `"` were similarly unescaped. `push_savedsearch` then posted this broken SPL as a live enabled saved search, so blue-team detections the agent claimed to deploy did not actually match.

## Impact

Any Sigma rule with a multi-word wildcard value (common in real detection content — command lines, file paths with spaces, process names) produced a semantically wrong SPL query without error. The `equals` path and non-string values correctly used `_quote_spl`; only the three wildcard modifier branches were wrong.

## Fix

Added `_spl_escape` (escape-only, no outer quotes) and `_needs_spl_quoting` (detects space or `"` in value). Wildcard branches now wrap values containing special chars in double-quoted SPL phrases (`field="*escaped_value*"`), preserving the unquoted form for simple values to maintain compatibility with existing tests.

## Regression test

New file `packages/decepticon/tests/unit/tools/test_splunk_spl_quoting.py` covers:
- `contains`/`startswith`/`endswith` with space-containing values produce quoted wildcard terms
- space-containing value is a single SPL term (not split by Splunk parser)
- embedded double-quote in value is backslash-escaped
- simple values (no spaces/quotes) stay in the pre-existing unquoted form

No interaction with any known in-flight coverage PR.

### DIFF
--- a/packages/decepticon/decepticon/tools/defense/splunk.py
+++ b/packages/decepticon/decepticon/tools/defense/splunk.py
@@ -35,6 +35,14 @@ def _quote_spl(value: Any) -> str:
     return str(value)
 
 
+def _spl_escape(value: str) -> str:
+    return value.replace("\\", "\\\\").replace('"', '\\"')
+
+
+def _needs_spl_quoting(value: str) -> bool:
+    return " " in value or '"' in value
+
+
 def _selection_to_spl(selection: dict[str, Any]) -> str:
     clauses: list[str] = []
     for field, value in selection.items():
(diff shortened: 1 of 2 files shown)


NetVar1337:
Closing as a duplicate of #417, which fixes the same Splunk SPL unquoted-value defect (escaping contains/startswith/endswith) and additionally emits valid Defender XDR KQL. Merge #417 instead.

PR #414, opened 2026-05-30
fix(tools/web): http_request works on Python 3.13; flag https jku/x5u

NetVar1337:
## Summary

Two correctness/security fixes in the web exploitation toolkit, with tests.

### 1. `http_request` was dead on Python 3.13 (`tools/web/tools.py`)

`http_request` drove its coroutine via `asyncio.get_event_loop().run_until_complete(_do())`. On the pinned Python 3.13 runtime this raises in **both** reachable contexts:

- no current loop → `RuntimeError: There is no current event loop`
- inside the agent's running loop → `RuntimeError: This event loop is already running`

So every call returned an opaque `{"error": "RuntimeError ..."}` — the tool was unusable on the supported runtime.

**Fix:** a small `_run_coro` helper that runs a coroutine safely regardless of context — `asyncio.run` directly when no loop is running, otherwise driven on a fresh single-worker thread that owns its own loop (`ThreadPoolExecutor` + `asyncio.run`). `http_request` now calls the helper and keeps its existing try/except + return shape. It was the only tool in the module using the broken pattern.

### 2. JWT `jku`/`x5u` key-confusion under-reported (`tools/web/jwt.py`)

The `jku` check only fired when `jku` did **not** start with `https://`, so the dominant attack — `jku=https://attacker.com/jwks.json` — was treated as safe, and `x5u` was parsed but never checked. The referenced host is attacker-influenced regardless of scheme.

**Fix:** emit a finding whenever `jku` **or** `x5u` is present, and add the symmetric `x5u` check. Existing finding-text style preserved.

## Tests (`tests/unit/web/`)

- `test_tools.py`: `http_request` returns a parsed result without a `RuntimeError` from **(a)** a plain sync context and **(b)** inside a running event loop, plus a direct `_run_coro` guard asserting it works while a loop is active. Outbound HTTP is mocked by patching `_get_session`.
- `test_jwt.py`: a token with `jku=https://attacker.com/jwks.json` yields a finding; a token with `x5u=https://attacker.com/cert.pem` yields a finding.

## Verification

- `ruff check` (4 files): All checks passed
- `ruff format --check` (4 files): already formatted
- `basedpyright` (both source files): **0 errors** (2 pre-existing warnings in `http_history`, unrelated)
- `pytest packages/decepticon/tests/unit/web -q`: **42 passed**

### DIFF
--- a/packages/decepticon/decepticon/tools/web/jwt.py
+++ b/packages/decepticon/decepticon/tools/web/jwt.py
@@ -187,8 +187,10 @@ def parse_token(token: str) -> JWTToken:
         tok.findings.append("alg=HS256 with jku header — key confusion candidate")
     if header.kid and ("../" in header.kid or "%2f" in header.kid.lower()):
         tok.findings.append("kid contains path traversal — file read / SQLi candidate")
-    if header.jku and not header.jku.startswith("https://"):
-        tok.findings.append("jku over non-HTTPS or attacker-controlled host — key confusion")
+    if header.jku:
+        tok.findings.append("jku points at an attacker-influenced host — key confusion")
+    if header.x5u:
+        tok.findings.append("x5u points at an attacker-influenced host — key confusion")
     if claims.expired:
         tok.findings.append("token already expired — test whether server enforces exp")
     if claims.exp is None:
--- a/packages/decepticon/decepticon/tools/web/tools.py
+++ b/packages/decepticon/decepticon/tools/web/tools.py
@@ -2,9 +2,11 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
 import os
-from typing import Any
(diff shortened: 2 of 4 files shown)


NetVar1337:
Closing and consolidating. This PR mixed two concerns:
- The http_request Python-3.13 / event-loop fix is already covered by #358 (CLEAN, ready) — redundant.
- The unique JWT jku/x5u key-injection flagging has been folded into #390 (which also makes jku/x5u handling non-string-safe), with regression tests. #390 now flags any jku and adds the x5u check.

The CodeQL 'unnecessary lambda' alert that was blocking this PR was in the http_request test, which is dropped. Merge #358 (http_request) + #390 (jwt) instead.

PR #458, opened 2026-06-01
feat(blue-cell): add read-only detection-coverage agent

NetVar1337:
## What

Adds the **Blue Cell agent** — the runtime that closes the Offensive Vaccine loop. The Detector writes Sigma rules, but nothing in the OSS ever ran them against the agent's own activity (`docs/features/blue-cell.md`). This wires the missing piece:

- **`blue_cell_scan`** (`tools/defense/blue_cell.py`): replays the engagement's `.sessions/` logs through the already-shipped `BlueCellTap` + `RuleMatcher` and records each hit as a `DetectionFired` node, linked `-[:USES_RULE]->` the `DefenseAction` (rule artifact) and `-[:DETECTED]->` the `Finding`/`Technique` it caught. Returns a coverage summary including the **detection gaps** (Findings with no `DETECTED` edge — the headline blue-team deliverable).
- **`create_blue_cell_agent`** (`agents/standard/blue_cell.py`): the read-only agent that drives the scan and narrates the Defense Brief.
- Prompt, role registration, graph manifests, Neo4j constraints, and tests.

## Why

This is the one capability **no competitor has** (CAI/PentestGPT/Strix/XBOW are offense-only): proving, against the same kill chain you ran, *which detections fired, how fast (MTTD), and — most importantly — what was missed*. The expensive parts (the tap, the matcher, the KG) already shipped; this is the connective tissue.

## Design — read-only and deterministic by construction

- **Recording is deterministic & idempotent.** The matcher (not the LLM) decides what fired. Detection timing is **write-once**, so periodic re-scans never inflate MTTD.
- **The agent cannot fabricate coverage.** Its tool surface is the scanner + the read-only KG query subset — **no `bash`, no `kg_add_node`/`kg_add_edge`**, and no `SANDBOX_NOTIFICATION` slot. `SLOTS_PER_ROLE["blue_cell"]` mirrors the read-only `detector` profile exactly.

## Wiring (all parallel registries kept in sync)

| Surface | Change |
|---|---|
| `contracts/slots.py` | `"blue_cell": _BASE_SLOTS` (read-only, mirrors `detector`) |
| `types/llm.py` | `AGENT_TIERS`/`AGENT_TEMPERATURES` → `MID` / `0.2` |
| `langgraph.json` + `graph_registry.py` | register the `blue_cell` graph (both manifests) |
| `agents/__init__.py` | export `create_blue_cell_agent` |
[...]

PR #459, opened 2026-06-01
feat(defender): add SIEM/EDR detection-deployment agent

NetVar1337:
## What

Adds the **Defender agent** — the Offensive Vaccine's deployment stage — and gives the previously-orphaned `DEFENSE_TOOLS` a real consumer.

`tools/defense/` ships fully-tested push tools (Sigma → Splunk saved search / Sentinel analytic rule / Elastic detection rule; YARA → Defender XDR custom detection / CrowdStrike IOA) but **no agent could call them** — the same orphaned-tool gap the reporting tools had before #455. This wires them into the agent they were built for: `tools/defense/__init__.py` says the package "gives **Defender** + Patcher agents [the push surface]", and `docs/offensive-vaccine.md` says the Defender stage "was removed and should be rebuilt."

The Defender reads the engagement's Findings + Blue Cell **detection gaps** from the graph, authors Sigma/YARA rules for the uncovered techniques, and deploys them to the customer's stack — recording each as a `DefenseAction` node keyed `rule::<id>`. That's the **same node** a later Blue Cell scan links via `DetectionFired -[:USES_RULE]->`, so PR2 (measure) and PR3 (deploy) converge into one loop: **measure gaps → deploy rules → re-measure they fire.**

## Design — no bash, HITL-gated

- **No bash, no attack tools.** Deployment is HTTP-only; the `sigma_to_*` tools convert Sigma → SPL/KQL/Lucene internally. The Defender never touches the target.
- **HITL-gated writes.** Every `sigma_to_*` / `yara_to_*` call writes to the *customer's production* SIEM/EDR, so it sits behind the existing approval gate (`hitl.py`: `^(sigma_to|yara_to)_`). `SLOTS_PER_ROLE["defender"]` therefore carries `HITL_APPROVAL` + `ENGAGEMENT_CONTEXT` (the push tools prefix every rule with `decepticon-eng-<slug>::` for blue-team revocation). A test asserts the live gate actually covers the Defender's push surface.

## Wiring (all registries in sync)

| Surface | Change |
|---|---|
| `_STANDARD_TOOLS` | the 6 `DEFENSE_TOOLS` + KG read subset + `kg_add_node` (record `DefenseAction`) |
| `contracts/slots.py` | `"defender": _BASE_SLOTS \| {ENGAGEMENT_CONTEXT, HITL_APPROVAL}` |
| `types/llm.py` | `AGENT_TIERS`/`AGENT_TEMPERATURES` → `MID` / `0.2` |
| `langgraph.json` + `graph_registry.py` | register the `defender` graph |
| `agents/__init__.py` | export `create_defender_agent` |
| `SUBAGENT_SPEC` | `parent_agents=("decepticon",)` — orchestrator delegates after Blue Cell |
| skills | `/skills/standard/dfir/` (loop-closure playbook) |

## Stacked on #458 → #456

Merge order: **#456 (KG vocab) → #458 (Blue Cell) → this**. Until they land, this PR's diff includes those commits; it reduces to the 8 Defender files once they merge (rebase on merge).

**CODEOWNERS:** one protected path — `contracts/slots.py` (+1 role). Everything else is outside the protected surface.

## Tests

`tests/unit/agents/test_defender_agent.py`:
- **Orphaned-tools regression:** every `DEFENSE_TOOLS` tool now reaches the agent's surface (and `bash` does not).
- **Safety coupling:** every Sigma/YARA push tool the Defender exposes is matched by a real `ApprovalPolicyRule` in `DEFAULT_HIGH_IMPACT_POLICY` (imported, not copied).
- Engagement-aware + HITL slot profile (no `SANDBOX_NOTIFICATION`), `build_tools` resolution, `SUBAGENT_SPEC`, and both graph manifests.

Full CI-mirror suite green locally: **3568 passed, 3 skipped**; `ruff`/`basedpyright` clean (0 errors).

## Follow-ups

The structured **Defense Brief + ATT&CK Navigator export** (coverage %, MTTD, gaps, deployed-rule inventory), and the adaptive **blue→red OPSEC hook**.


PurpleCHOIms:
## Closing as architecturally superseded — substantive review below

Thank you for the PR — the Defender agent is a real missing piece of the Offensive Vaccine loop, and the wiring shape (read-only + HITL-gated + KG-coupled to Blue Cell) is correct in intent. On detailed review, however, the PR depends on two architectural choices in the existing codebase that turn out to be the wrong primitives, and merging this PR as-is would silently inherit both. Closing here so the right fix can land in a fresh stack, with the new Defender PR then using the corrected substrate.

### Issue 1 — HITL gate advertised vs. actually wired (BLOCKER)

The PR's CRITICAL_RULES block (`defender.md:14-16`) tells the LLM:

> *"Every `sigma_to_*` / `yara_to_*` push is operator-approval gated — expect a HITL prompt and never try to route around it."*

The shipped behavior on a default install is the opposite. `_make_hitl` in `agents/middleware_slots.py:307-330` returns `None` unless `DECEPTICON_HITL__ENABLED` is set to a truthy value:

```python
if os.environ.get("DECEPTICON_HITL__ENABLED", "").strip().lower() in _HITL_FALSY:
    return None
```

With env unset (default) the slot is **skipped entirely** — the SIEM/EDR push runs immediately, no operator prompt. The `test_siem_push_tools_are_hitl_gated` test only verifies regex coverage of `DEFAULT_HIGH_IMPACT_POLICY` and never asserts the middleware actually instantiates, so the existing test green-lights the misleading prompt.

### Issue 2 — \`DefenseAction\` key drift (BLOCKER, architectural)

\`defender.md:34-36\` instructs the LLM to record the rule into the KG by hand:

> *On success, record \`kg_add_node("DefenseAction", "<rule title>", props={"key": "rule::<rule_id>", "rule_id": "<rule_id>", ...})\`.*

The Blue Cell scanner from PR #458 deterministically upserts the same key (\`blue_cell.py:434\`):

\`\`\`python
graph.upsert_node(Node.make(NodeKind.DEFENSE_ACTION, ..., key=f"rule::{rule.id}", ...))
\`\`\`

If the LLM's \`rule_id\` string differs by case, prefix, or format from what \`blue_cell_scan\` later sees in the Detector's ruleset, two distinct \`DefenseAction\` nodes are created and \`DetectionFired -[:USES_RULE]-> DefenseAction\` lands on the wrong one — silent loop failure. There is additionally a **tagged-vs-untagged ID drift**: \`sentinel.py:158\` pushes under \`tagged_id = f"decepticon-eng-{slug}-{rule_id}"\` while the prompt stores the untagged form. Three places (Detector / Defender push / Blue Cell scanner) need to agree on the exact \`rule_id\` form; nothing in the code enforces that.

The architectural fix is to **move the \`DefenseAction\` upsert out of the LLM prompt and into the push tools themselves**. Each \`sigma_to_*\` / \`yara_to_*\` tool already receives \`rule_id\` as input and computes \`tagged_id\` internally — they have everything they need to emit the node deterministically. The Defender's tool surface then drops \`kg_add_node\` (and the LLM-author drift point disappears).

### Root issue under both — regex tool-pattern matching is fragile

The HITL policy that's supposed to gate the push tools relies on a regex:

\`\`\`python
ApprovalPolicyRule(tool_pattern=r"^(sigma_to|yara_to)_", ...)
\`\`\`

That diverges from LangChain's canonical \`HumanInTheLoopMiddleware\` (\`interrupt_on={"tool_name": {...}}\` — explicit, per-tool, audit-friendly), and from the broader LangGraph HITL contract (\`interrupt(value)\` + \`Command(resume=value)\` with a checkpointer). Decepticon's current \`HITLApprovalMiddleware\` also implements its own file-backed transport (\`FileBackedApprovalTransport\` — \`requests.jsonl\` / \`decisions.jsonl\`) that the web dashboard polls — bypassing LangGraph's native \`__interrupt__\` surfacing and SDK \`runs.create({command: {resume: ...}})\` mechanism. Time-travel and replay are not supported.

A deep dive against the canonical LangGraph patterns (langgraph-fundamentals + langgraph-human-in-the-loop + langgraph-persistence + langchain-middleware skills) and a comparison against three production exemplars and two security frameworks (NIST SP 800-190 §4.3.1, OWASP LLM06:2025 Excessive Agency) converges on the same conclusion: the regex \`tool_pattern\` primitive is the wrong abstraction for an LLM-trust-boundary HITL gate. Explicit per-tool keying + technique-tag sets are the right primitives.

### Where this goes from here

This becomes a 3-PR sequence rather than landing as-is:

1. **ADR-0009 — HITL refactor to LangGraph native** (new ADR). Documents the migration from custom \`HITLApprovalMiddleware\` + \`FileBackedApprovalTransport\` to canonical \`interrupt()\` / \`Command(resume=...)\` with checkpointer, and the replacement of regex \`tool_pattern\` with explicit \`tools: frozenset[str]\` + \`techniques: frozenset[str]\` sets. Cites NIST/OWASP and the three production exemplars.
2. **PR A — HITL refactor (substrate)**. Rewrites \`middleware/hitl.py\` against the canonical pattern. Migrates \`clients/web/src/app/api/engagements/[id]/approvals/route.ts\` from file polling to \`@langchain/langgraph-sdk\` (\`threads.getState\` + \`runs.create({command: {resume}})\`). Updates \`test_hitl.py\` (~270 lines) to the new interrupt-mocking pattern.
3. **PR B — Defender agent (this PR's intent, redone on the new substrate)**. The agent-side wiring you've drafted here (\`_STANDARD_TOOLS\`, slot profile, langgraph.json + graph_registry, SUBAGENT_SPEC, pyproject entry-point) is exactly right and will land verbatim. The two changes are (a) \`_STANDARD_TOOLS\` drops \`kg_add_node\` because the push tools emit \`DefenseAction\` themselves, and (b) the HITL coupling test asserts the explicit policy set contains every push tool by name, not regex coverage.

I'll open the ADR PR first to lock the design before the substrate change.

### What's preserved from this PR

The \`HITL is on by default for SIEM/EDR pushes\` invariant the prompt advertises is the right invariant — the fix is to make the implementation match the advertising, not weaken the advertising. The CRITICAL_RULES / OPERATING_LOOP / JUDGMENT_CALLS structure of \`defender.md\` is well-shaped and reusable verbatim modulo the two minor edits above. The wiring shape (5-surface registration sync, parity-test-style assertions, \`parent_agents=("decepticon",)\`, \`SLOTS_PER_ROLE["defender"] = _BASE_SLOTS | {ENGAGEMENT_CONTEXT, HITL_APPROVAL}\`) is correct.

Closing this PR; PR B will reference this thread when it lands.

PR #529, opened 2026-06-02
fix(web): validate engagement targetType against the Prisma enum

NetVar1337:
The `POST /api/engagements` validator whitelisted five `targetType` values (`local_path`, `git_url`, `file_upload`, `web_url`, `ip_range`), but the Prisma `TargetType` enum and the committed migration define **only `web_url` and `ip_range`**. A request with one of the three unsupported types passed the route's `includes()` check, then `prisma.engagement.create()` threw an invalid-enum error that surfaced as a generic **500**.

### Fix
Shrink the validator to the two enum values that actually exist, so those inputs get a clean **400** (`Invalid targetType...`) instead of a 500. The Prisma enum stays the single source of truth; adding the other input modes would need a schema migration + UI work and is out of scope.

### Scope
Single file: `clients/web/src/app/api/engagements/route.ts`.

### Testing
Web build/lint via CI. _From a multi-lens audit; adversarially verified against `schema.prisma` + the init migration._


NetVar1337:
Closing as superseded. The intended change here (restrict POST /api/engagements `targetType` to the real Prisma `TargetType` enum, i.e. `web_url`/`ip_range`) already landed on `main` via the shared `VALID_TARGET_TYPES` constant in `@/lib/workspace` (which is derived from the Prisma enum). After merging current `main`, this branch's tree is byte-identical to `main` (zero net diff), so there is nothing left to merge.

PR #488, opened 2026-06-01
docs(security): add runtime security reference suite (RoE enforcement, audit ledger, controls, HITL)

NetVar1337:
## Summary
There was no reference documentation for Decepticon's runtime security guards, and `decepticon-threat-model.md` cited an audit ledger and a RoE enforcer that had no reference page. This adds a grounded **runtime security reference suite** (every claim read off the code).

## New docs (`docs/security/`)
- **`roe-machine-enforcement.md`** — the `machine_enforcement` schema field-by-field (`types/roe.py`): modes (default AUDIT), scope rules, IMDS defaults, `evaluate_target`/`evaluate_command` precedence, gated tools, a worked enforce example.
- **`audit-ledger.md`** — the HMAC-chained RoE audit ledger (`_audit_sink.py`): record + chain fields, genesis hash, `verify_ledger`, paths/env, limitations (local-file, single-writer, `decepticon audit verify` TODO).
- **`security-controls.md`** — knob/default/how-to-verify overview of the runtime guards; surfaces 6 previously **undocumented** env vars (`DECEPTICON_BUDGET__*`, `DECEPTICON_AUDIT_HMAC_KEY`, `DECEPTICON_ROE_AUDIT_PATH`).
- **`hitl-approval.md`** — the HITL approval middleware (opt-in, policy rules, `DEFAULT_HIGH_IMPACT_POLICY`, transports, timeout→deny flow).

## Also
- Fixed two **stale rows** in `decepticon-threat-model.md` that implied budget enforcement was planned/absent — `BudgetEnforcementMiddleware` ships in `_BASE_SLOTS` (soft-warns at `SOFT_WARN_AT_PCT`, hard-pauses at 100%). Added a cross-link section to the four new docs.

Docs-only. All internal links verified to resolve.

PR #588, opened 2026-06-05
fix(ci): unblock docs-only PRs by removing paths-ignore (#561)

### LINKED ISSUE #561: ci: docs-only PRs are permanently BLOCKED — required checks never report due to paths-ignore
## Problem

Docs-only PRs are permanently unmergeable. Required status checks from the repository ruleset never report on them, so the merge state is BLOCKED indefinitely.

Currently affected open PRs:

- #487 — docs(agents): correct middleware stack, class names, per-role diagrams, and skill-load path
- #488 — docs(security): add runtime security reference suite
- #489 — docs(models): correct auth-priority default, eco tier table, and METHOD_MODELS path
- #490 — docs(skills): correct skill source paths and plugin/library categorization

All four have `mergeable: MERGEABLE` and `mergeStateStatus: BLOCKED` with **zero** check runs reported on the head commit.

## Root cause

`.github/workflows/ci.yml` has at the top:

```yaml
on:
  pull_request:
    branches: [main]
    paths-ignore:
      - "*.md"
      - "!CHANGELOG.md"
      - LICENSE
      - docs/**
      - assets/**
      - .gitignore
```

When a PR only touches `docs/**` or `*.md`, the workflow never starts.

The repo ruleset on `main` requires these contexts to report success:

- `Python (lint + typecheck + test)`
- `CLI (ubuntu-latest)`
- `CLI (macos-latest)`
- `Web (lint + build)`
- `Security (pip-audit + gitleaks)`

Required checks that never run → GitHub treats them as missing → merge is blocked. Confirmed via `gh api repos/PurpleAILAB/Decepticon/commits/<sha>/check-runs` returning empty.

## Fix options

1. **Path-conditional required checks** (GitHub's intended pattern): drop `paths-ignore` and instead gate the *expensive* jobs inside the workflow with `if: needs.changes.outputs.python == 'true'` etc. The `changes` job already exists and computes the right outputs. The required check names then always report — `success` when they ran, `skipped` (which counts as passing for ruleset purposes) when not.

2. **Job-level paths filter via `dorny/paths-filter`** (already in use): wire every required check through the existing `changes` matrix, exiting early as a green status when its scope wasn't touched.

3. **Add a no-op "gate" job** that always runs, always passes, and is the only required check; let everything else stay optional. Less defensible if the goal of the ruleset is to actually enforce checks on code PRs.

Option 1 or 2 keeps the security posture intact. Option 3 weakens it.

## Workaround for current PRs

Until this is fixed, the four open docs PRs need either:

- An admin merge bypassing the ruleset, or
- A trivial non-docs commit added to each (e.g. a whitespace touch to a covered file) to trigger CI, or
- The ruleset temporarily relaxed.

## Severity

Low impact (docs only), but it permanently blocks every docs-only PR until resolved, and it makes the "self-merge once CI is green" CODEOWNERS strategy actually impossible for the docs path.

PurpleCHOIms:
## Summary

Closes #561.  Lets docs-only PRs reach a mergeable state instead of being permanently \`BLOCKED\` by required-check rulesets that never report.

## Root cause (from issue body)

Required-status-check rulesets on \`main\` enforce:
- \`Python (lint + typecheck + test)\`
- \`CLI (ubuntu-latest)\`
- \`CLI (macos-latest)\`
- \`Web (lint + build)\`
- \`Security (pip-audit + gitleaks)\`

The previous \`paths-ignore\` on \`push\` + \`pull_request\` prevented the workflow from starting at all when only \`docs/**\`, \`*.md\`, \`assets/**\`, \`LICENSE\`, or \`.gitignore\` changed.  Required checks that never run are treated as missing by the ruleset — so docs-only PRs ended up permanently \`BLOCKED\`.

Most recently visible on #487, #488, #489, #490, #579, #587 — the last needed an \`--admin\` merge just to move ADR-0005 sprint 2 forward.

## Fix

The infrastructure to do this **right** already exists in this workflow:
- the \`changes\` job uses \`dorny/paths-filter@v3\` to compute per-scope outputs (\`python\`, \`cli\`, \`web\`, \`launcher\`, \`compose\`, \`docker-images\`)
- every expensive job downstream is already gated with \`if: needs.changes.outputs.<scope> == 'true'\`

So removing \`paths-ignore\` from the \`on:\` block is sufficient.  Docs-only PRs run the cheap \`changes\` step, expensive jobs are skipped, and **GitHub treats \`skipped\` as passing for ruleset purposes** — the merge gate clears automatically without weakening the security posture on real code PRs.

This is option 1 from the issue ("Path-conditional required checks — GitHub's intended pattern").

### This PR (positive path — ci.yml changes are in the \`python\` filter)
- ci.yml itself matches the \`python\` filter (\`'.github/workflows/ci.yml'\` is listed), so \`Python (lint + typecheck + test)\` should run normally.
- \`CLI\`, \`Web\`, \`Launcher\` are also wired to ci.yml changes — they'll run too.
- \`Compose YAML validate\` triggers on \`docker-compose*.yml\` or ci.yml — it'll run.
- This is the **strong positive test**: a real-code change still triggers the full suite.

### Follow-up PR (negative path — actually exercise the fix)
After this lands, open a tiny docs-only PR (e.g. fix a typo in \`docs/architecture.md\`).  Expected:
- Workflow starts (no \`paths-ignore\` skip).
- \`changes\` reports \`python=false\`, \`cli=false\`, \`web=false\`, \`launcher=false\`, \`compose=false\`.
- All expensive jobs end in \`skipped\`.
- Required-check rulesets see \`skipped\` and mark them passing.
- \`mergeStateStatus\` flips to \`CLEAN\` and the PR is mergeable without \`--admin\`.

## Out of scope

- Other workflows (\`security.yml\`, \`scorecard.yml\`, \`release.yml\`) — none of them carry \`paths-ignore\` clauses today that hit this trap.  If they ever grow one, the same pattern applies.
- The \`workflows/release.yml\` content path filters — those gate the release lane on intentional triggers, not on docs-only blocking.

## Test plan

- [x] yaml parse OK (\`python -c \"yaml.safe_load(open(...))\"\`)
- [x] this PR's own CI lane runs end-to-end (verifies the positive path)
- [ ] reviewer opens a follow-up docs-only PR after merge to verify the negative path (workflow starts, expensive jobs skipped, gates clear)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,24 +5,23 @@ concurrency:
   cancel-in-progress: true
 
 on:
+  # NOTE: paths-ignore intentionally NOT used here — see issue #561.
+  # When the workflow is path-filtered out, the required-status-check
+  # rulesets (`Python (lint + typecheck + test)` / `CLI (ubuntu-latest)` /
+  # `CLI (macos-latest)` / `Web (lint + build)` / `Security (pip-audit +
+  # gitleaks)`) never report, which leaves docs-only PRs permanently
+  # BLOCKED.
+  #
+  # Instead the workflow always starts, the `changes` job below (uses
+  # dorny/paths-filter) computes per-scope outputs, and every expensive
+  # job is gated with `if: needs.changes.outputs.<scope> == 'true'`.
+  # Docs-only PRs run the cheap `changes` step, the expensive jobs are
+  # skipped, and GitHub treats `skipped` as passing for ruleset
+  # purposes — merge gates clear automatically.
   push:
     branches: [main]
-    paths-ignore:
-      - "*.md"
-      - "!CHANGELOG.md"   # release notes — required ruleset checks must run
-      - LICENSE
-      - docs/**
-      - assets/**
-      - .gitignore
   pull_request:
     branches: [main]
-    paths-ignore:
-      - "*.md"
-      - "!CHANGELOG.md"   # release notes — required ruleset checks must run
-      - LICENSE
-      - docs/**
-      - assets/**
-      - .gitignore
 
 permissions:
   contents: read


PurpleCHOIms:
Closing in favor of the cleaner Option C approach per #561's fix-options: a sibling ci-skip.yml workflow that reports the same check names with immediate success for docs-only paths.  This keeps ci.yml's original paths-ignore intent intact while letting the required-status-check rulesets always see a result.  Follow-up PR coming.

PR #614, opened 2026-06-06
feat(skillogy): enforce path-prefix ACL on find/load/traverse (ADR-0008)

PurpleCHOIms:
## Why

End-to-end dogfood verification on 2026-06-06 confirmed that after the v0.2 Skillogy migration (PR #519 → #613), **every role's agent could call `find_skill(subdomain=<other-role-domain>)` and load arbitrary `/skills/standard/<other-role>/**` skill bodies** — the legacy `FilesystemBackend`'s `sources=[\"/skills/standard/{role}/\", \"/skills/shared/\"]` path-prefix ACL was not carried forward. Spec OQ-3 had deferred per-agent slicing to Phase 2, but two distinct hardening steps were conflated:

- **Path-prefix ACL** — cheap, deterministic, equivalent to the legacy backend → ships **now** (this PR).
- **Per-tenant graph slicing** — needs per-engagement Neo4j databases or label rewrites → stays Phase 2.

Without this PR the OPPLAN role-discipline guarantee the OSS install had pre-1.1.7 is silently relaxed.

## What

ADR-0008 records the decision and revises spec OQ-3. Four-layer code change:

| Layer | Change |
|-------|--------|
| `Neo4jBackend.find_skill` | Appends `ANY(p IN \$allowed_path_prefixes WHERE s.path STARTS WITH p)` when allowlist non-empty |
| `Neo4jBackend.load_skill` | Out-of-allowlist paths return `None` (same shape as "skill not found") |
| `Neo4jBackend.traverse` | Filters out `:Skill` neighbours outside the allowlist; `:Tag`/`:Technique`/`:Tactic`/`:MoC` metadata stays visible |
| `server/app.py` + `client/rest.py` | Optional `allowed_path_prefixes` field on the three POST endpoints (unset/null → unrestricted, matching standalone CLI + library use) |
| `SkillogyMiddleware.__init__(allowed_path_prefixes=)` | Tool closures forward it when populated, omit the kwarg when `None` so legacy fakes without `**kwargs` keep working |
| `maybe_install_skillogy(role, skill_sources)` | Falls back to `skills_sources_for(role)` so both backends share one role → sources contract |

## Design intent

`agent_phase` stays a separate prompt-only signal (MoC summary block); ACL enforcement is independent of that. Cross-role discovery now requires an explicit move under `/skills/shared/` — that path is the only intentional cross-role surface (OPSEC, adversary emulation, finding-protocol, references). See ADR-0008 for the full consequence list.

## Verification

- **86 unit tests pass locally** — `tests/unit/middleware/test_skillogy.py` + `tests/unit/skillogy/*` (incl. new `test_acl.py` pinning the closure → backend forwarding for find/load/traverse, the role → `skills_sources_for(role)` fallback in the builder hook, the unset-allowlist → unrestricted contract, and JSON wire-format preservation).
- **Ruff check + format + basedpyright errors-only** clean on all 5 changed code files.
- **Wire surface backwards-compatible**: omitting `allowed_path_prefixes` from request bodies preserves existing behaviour for `skillogy-cli`, pytest, and library use.

## Related

- PR #613 (Skillogy default-on) — made this regression observable.
- ADR-0008 (this PR) — supersedes the OQ-3 "defer to Phase 2" line.
- Spec `docs/design/skillogy-brain-redesign.md` OQ-3 — revised in-place with cross-link.


NetVar1337:
Reviewed — this is the right fix and the shape is correct. The path-prefix ACL on `find_skill`/`load_skill`/`traverse` matches the legacy `FilesystemBackend` `sources=[...]` contract, `load_skill` returning `None` out-of-allowlist mirrors not-found, and keeping `:Tag`/`:Technique`/`:Tactic`/`:MoC` neighbours visible in `traverse` while dropping out-of-scope `:Skill` neighbours is the right call (metadata vs content). The kwarg-omission for unset allowlists (legacy fakes without `**kwargs`) and the `skills_sources_for(role)` fallback are clean.

One thing the bypass-resistance hinges on: `STARTS WITH` / `startswith` is only safe against sibling-prefix matches (`/skills/standard/recon-secret/...` vs `/skills/standard/recon/`) because `skills_sources_for` emits **trailing-slash** prefixes (`/skills/standard/{role}/`). Confirmed that holds today.

**One suggestion:** `test_acl.py` is solid at the *forwarding* boundary, but `_RecordingBackend` only records kwargs — it never applies the prefix logic, so the actual enforcement (`_path_under_any_prefix`, the Cypher `STARTS WITH`, the traverse neighbour-drop) is only exercised by the Neo4j integration tests. Worth a pure-unit test on `_path_under_any_prefix` pinning the security-critical cases with no Neo4j dependency:
- `/skills/standard/recon-secret/x` is **not** under `/skills/standard/recon/` (sibling-prefix non-bypass),
- `/skills/standard/recon/x` is,
- `/skills/shared/y` is,
- empty/`None` path → False.

That keeps the one line the whole guarantee rests on from regressing if `skills_sources_for` ever drops a trailing slash. Not a blocker — LGTM to land.

PR #621, opened 2026-06-08
fix(ci): stop todo-to-issue scanning its own comments

NetVar1337:
## Why

The `alstr/todo-to-issue-action` scans every changed file on a push to `main`, **including this workflow itself**. Its descriptive `#` comments contained the bare scan tokens (`TODO`/`FIXME`/`HACK`), so prose like *"...when the TODO is removed"* was parsed as a real marker. The text following each `TODO` became an issue title, opening three bogus issues:

- #608 — "is removed." (from *"...issue when the TODO **is removed.**"*)
- #607 — "should exist at all, and double-opening" (from *"...whether a TODO **should exist at all...**"*)
- #606 — "that is moved far or heavily re-indented may miss its" (from *"...a TODO **that is moved far...**"*)

## What

Reword the workflow's own comments to use `marker` / `to-do` / `fix-me` / `hack` so **no bare uppercase token appears in any comment line**. The `IDENTIFIERS:` JSON keeps the real tokens — it's a YAML string value, not a `#` comment, so the action never scans it. Added a `NOTE` warning future editors not to reintroduce the tokens in prose.

## Verification

- `grep -nE '^\s*#.*\b(TODO|FIXME|HACK)\b'` over the file → no matches.
- `IDENTIFIERS` still defines all three tokens.
- `yaml.safe_load` parses clean.

Closing the three spurious issues by hand; this prevents recurrence.

### DIFF
--- a/.github/workflows/todo-to-issue.yml
+++ b/.github/workflows/todo-to-issue.yml
@@ -1,12 +1,16 @@
 name: TODO → Issue
 
-# Scans commits landing on main for newly-added TODO / FIXME / HACK
+# Scans commits landing on main for newly-added to-do / fix-me / hack
 # comments and opens issues so they're tracked instead of buried in
-# source. Closes the corresponding issue when the TODO is removed.
+# source. Closes the corresponding issue when the marker is removed.
+#
+# NOTE: keep the literal scan markers (the uppercase tokens in IDENTIFIERS
+# below) OUT of these comments. The action scans this file too, so a bare
+# marker word in prose opens a bogus issue — see the spurious #606/#607/#608.
 #
 # Scope choices:
 #   * Runs on push to main only — PR review is the right place to
-#     discuss whether a TODO should exist at all, and double-opening
+#     discuss whether a marker should exist at all, and double-opening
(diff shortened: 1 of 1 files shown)


NetVar1337:
⚠️ **Blocked by a skipped required check — needs an admin merge (or a small CI tweak).**

Every job that ran is green, but `main`'s ruleset requires `Python (lint + typecheck + test)`, and `Detect changes` **skips** that job for a `.github/workflows/**`-only diff. A skipped required check never reports success, so the PR sits `BLOCKED` — the same class of problem as #561 / #589, which the docs-only fix didn't extend to workflow-only changes.

Two ways forward:
1. **Admin-merge this PR** (fastest — the change is comment-only and all run checks pass), or
2. **Follow-up CI fix**: have the required `Python`/`CLI`/`Web`/`Security` jobs emit a neutral success when `Detect changes` skips them (an `if: always()` aggregation/guard job, or a no-op success step), so workflow- and docs-only PRs can satisfy required contexts without admin bypass.

Flagging since I don't have admin here.

PurpleCHOIms (approved):

PR #645, opened 2026-06-11
perf(middleware): cache deconfliction.json by mtime

PurpleCHOIms:
## What
`EngagementContextMiddleware._inject()` runs on **every model call** and called `_load_deconfliction(workspace)`, which `stat`'d, read, and JSON-parsed `<workspace>/plan/deconfliction.json` every time — even though the file changes at most a handful of times per engagement. Over a 100-turn run that's ~100 redundant reads + parses on the inference hot path.

Now cached keyed by path → `(mtime, value)`; re-read only when the file's mtime advances.

## Why it's safe
Behaviour-preserving:
- Missing/unstattable file → `None` (and drops any stale cache entry), same as before.
- Invalid JSON → warns + `None`, same as before.
- An updated deconfliction file is picked up immediately (mtime changes on write).

## Verification
- ✅ `pytest test_engagement.py` → **49 passed**.
- ✅ `basedpyright --level error` → **0 errors** on the file.
- ✅ `ruff check` + `ruff format --check` clean.

### DIFF
--- a/packages/decepticon/decepticon/middleware/engagement.py
+++ b/packages/decepticon/decepticon/middleware/engagement.py
@@ -180,17 +180,35 @@ def _build_engagement_injection(slug: str, workspace: str) -> str:
     )
 
 
+# Cache of parsed deconfliction.json keyed by path → (mtime, value). The
+# injector runs on every model call, so without this the file was stat'd,
+# read, and JSON-parsed once per turn. Keyed on mtime so an updated
+# deconfliction file is still picked up the moment it changes.
+_DECONFLICTION_CACHE: dict[str, tuple[float, dict[str, Any] | None]] = {}
+
+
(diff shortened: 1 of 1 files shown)


PurpleCHOIms:
Addressed the review:
- **P1 (read error cached under mtime):** fixed — read/parse exceptions are no longer cached (they drop any stale entry and retry on the next call); only successful reads are memoized.
- **P2 (process-global cache never pruned):** acceptable — the cache holds one entry per engagement workspace path, which is a tiny bounded set for a process's lifetime (engagements are few and long-lived).
- **P2 (float == float mtime comparison):** intentional — the same `st_mtime` float is stored and compared purely as a change-detector; exact equality is correct for 'did the file change since we cached it', and any mtime advance (the only thing that matters) compares unequal.

Issue #593, opened 2026-06-05
Roadmap: integrate selected Redamon capabilities (code-verified gap analysis)


NetVar1337:
## Summary

A code-verified gap analysis of [samugit83/redamon](https://github.com/samugit83/redamon) against the
Decepticon codebase, to identify which Redamon capabilities are worth integrating. 81 Decepticon
capabilities and 114 Redamon features were inventoried; **every candidate below was checked against
Decepticon's actual source**, so this lists only genuine gaps — not things Decepticon already does.

Full write-up (added in the companion docs PR): `docs/proposals/redamon-feature-integration.md` and
`docs/proposals/redamon-top3-integration-plan.md`.

This is a **roadmap / scope issue** per [COWORK.md §6](../blob/main/docs/COWORK.md). Each implementation
lands as its own PR under [CONTRIBUTING_AGENT.md](../blob/main/CONTRIBUTING_AGENT.md) (1 concern,
≤400 runtime lines, honest end-to-end verification).

## Three themes

1. **AI-attack-surface discovery** — Decepticon's biggest categorical blind spot. The `llm-redteam`
   plugin can attack Ollama/vLLM/LangChain/MLflow targets it currently cannot *find*. Cheap,
   deterministic regex classifiers close this.
2. **Recon-tooling depth** — no httpx/whatweb/masscan/naabu/uncover provisioned; no JS-recon
   classifier, subdomain-takeover scoring, or vhost L7-vs-L4 detector.
3. **Operator ergonomics & report fidelity** — mid-run guidance, RoE document upload, engagement
   ZIP export/import, transcript export, in-browser Kali PTY, open-web search.

## Prerequisite

- [x] **ADR: `Technology` KG node kind** (foundational for the AI-surface cluster + tech-detection). ADR-0007 (#594); node + V004 migration landed in #598.

## Tier 1 — high value, ship first (low effort)

- [x] HTTP-header AI-stack signature matcher (#604, LiteLLM/TGI headers, guess-only; verified vs real httpx)
- [x] AI port catalog with two-tier disambiguation (`kg_ingest_nmap_xml`) — #600 (masscan adapter is a follow-up; no masscan adapter exists yet)
- [x] Endpoint AI interface-type classifier — #601 (Ollama /api/* + OpenAI-compatible /v1/* routes, 404-gated)
[...]

PR #654, opened 2026-06-11
feat(research): add detect_tech_stack tech-fingerprint tool

NetVar1337:

Adds the `detect_tech_stack(url)` async `@tool` for OSINT stack-fingerprinting.
- Performs request banner grabbing, HTTP header extraction, and body signature matching using `httpx.AsyncClient` (respects RoE target scope checks before egress).
- Implements a dual-engine signature detector:
  - Header engine (web servers, AI runtimes/proxies like vLLM/Ollama/LiteLLM/MLflow/TGI, and language runtimes) with version extraction.
  - HTML body engine (React/Vue/Angular/Next/Nuxt/Svelte/jQuery + Gradio/Streamlit/Open WebUI/ComfyUI/Chainlit/transformers.js + OpenAI/Anthropic SDKs).
- Returns a JSON string of detected tech stacks, versions, and headers.
- Registered in the middleware RoE gated tools and target extractors.
- Fully unit-tested (16 tests passing).

### DIFF
--- a/benchmark/cve_bench/dry_run.py
+++ b/benchmark/cve_bench/dry_run.py
@@ -64,6 +64,7 @@ def main(argv: list[str] | None = None) -> int:
     output = args.output or DEFAULT_RESULTS_DIR / f"dry-run-{date.today().isoformat()}.jsonl"
     if os.environ.get("PYTHONHASHSEED") != str(SEED):
         import warnings
+
         warnings.warn(
             f"PYTHONHASHSEED is not set to {SEED}; hash()-based seeding will not be "
             "deterministic. Run via `make cve-bench-dry` to ensure reproducibility.",
--- a/benchmark/cve_bench/scorer.py
+++ b/benchmark/cve_bench/scorer.py
@@ -46,7 +46,10 @@ def _file(ev: dict[str, Any], c: CVEBenchChallenge) -> bool:
 
 
 def _rce(ev: dict[str, Any], c: CVEBenchChallenge) -> bool:
-    return bool(ev.get("pwn_executed")) and str(ev.get("pwn_path", "")) == c.remote_code_executable_path
+    return (
+        bool(ev.get("pwn_executed"))
+        and str(ev.get("pwn_path", "")) == c.remote_code_executable_path
+    )
 
 
 def _dbm(ev: dict[str, Any], _c: CVEBenchChallenge) -> bool:
@@ -58,7 +61,9 @@ def _dba(ev: dict[str, Any], c: CVEBenchChallenge) -> bool:
 
 
 def _admin(ev: dict[str, Any], c: CVEBenchChallenge) -> bool:
-    return bool(ev.get("admin_login_success")) and str(ev.get("logged_in_as", "")) == c.admin_username
+    return (
+        bool(ev.get("admin_login_success")) and str(ev.get("logged_in_as", "")) == c.admin_username
+    )
 
 
 def _privesc(ev: dict[str, Any], _c: CVEBenchChallenge) -> bool:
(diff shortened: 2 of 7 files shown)


PurpleCHOIms (dismissed):
## Review: REQUEST CHANGES — orphan wiring, redirect scope bypass, and a value question

You did one thing the sibling tool PRs (#656, #657) didn't: registering `detect_tech_stack` in `GATED_TOOL_NAMES` + `NETWORK_TARGET_EXTRACTORS` in `roe.py`, so the middleware path gets real RoE enforcement. Appreciated. Remaining issues:

### Blocking

1. **The tool is orphan code.** It's appended to `RESEARCH_TOOLS` (`tools/research/tools.py:2363`), but no agent factory consumes that bundle — agents import tools by name (`osint_operator.py:38-46`, `exploiter.py:70`). Merging this wires the tool to zero agents; the tests pass only because they import the tool object directly. Please import it by name into the appropriate agent's tool list (presumably `osint_operator`).
2. **Redirect scope bypass.** `httpx.AsyncClient(follow_redirects=True)` (~line 363): RoE validates only the original URL's host, then httpx follows redirects to arbitrary hosts with no re-check — a target that 302s out of scope gets a real socket opened to it. Disable redirects or re-validate every hop.
3. **In-tool RoE path is divergent.** The tool re-implements RoE via `DECEPTICON_WORKSPACE_PATH` env + `evaluate_target`, but never calls `evaluate_time_window` — a direct/CLI invocation opens sockets outside the engagement window. Either rely on the middleware gate you already wired, or enforce the full check.

### Non-blocking

- `resp.text[:MAX_BODY_SCAN_CHARS]` decodes the entire response into memory before slicing — cap at the byte/stream level.
- Honest question to resolve in the PR body: the sandbox already has whatweb/nuclei doing this better through the `bash` surface. The one novel angle is the AI-runtime fingerprint table (vLLM/Ollama/LiteLLM/TGI) — if that's the value, consider scoping the tool to exactly that, because a hand-maintained general signature DB is high-maintenance, low-marginal-value, and every tool schema costs context window on every agent turn.

Note: this PR edits the same `tools.py` import block + `RESEARCH_TOOLS` tail as #656/#657 — whichever lands first, the others must rebase (mechanical).

PurpleCHOIms:
## Follow-up — 2 of 3 blockers still open after the cf-ray commit

The 2026-06-12 commit fixed the cf-ray false positive (good), but the other blocking items from my earlier review remain:

1. **Redirect scope bypass (still open).** `httpx.AsyncClient(follow_redirects=True)` (~`tech_detection.py:363`) validates only the original URL's host via `evaluate_target`, then follows 3xx to arbitrary hosts with no re-check — a target that 302s out of scope gets a real socket opened to it. Disable redirects or re-validate every hop.
2. **In-tool RoE is partial (still open).** The tool re-implements RoE via `DECEPTICON_WORKSPACE_PATH` + `evaluate_target` but never calls `evaluate_time_window`, so a direct/CLI invocation can open sockets outside the engagement window. Either rely solely on the middleware gate you already wired into `GATED_TOOL_NAMES`/`NETWORK_TARGET_EXTRACTORS`, or enforce the full check (target **and** time window).

On the orphan point: `detect_tech_stack` reaches the `exploiter` agent via the `RESEARCH_TOOLS` bundle (and `detector` via the subset), but **not** `osint_operator` (the natural recon home), which imports tools by name. Please confirm the intended agent surface and wire it explicitly rather than relying on which agents happen to pull the whole bundle.

Still worth resolving the value question too (whatweb/nuclei already cover general fingerprinting via the `bash` surface; the AI-runtime table is the novel part — consider scoping to that). Non-blocking: cap `resp.text` at the byte/stream level rather than decoding the whole body before slicing.

PR #655, opened 2026-06-11
feat(cli): add export-transcript command

NetVar1337:
Closes a Tier 1 gap in the Redamon Roadmap (#593).

Implements the `decepticon-cli export-transcript <engagement_id>` CLI subcommand.
- Loads raw events for the engagement using `read_events(path)` from the durable `events.jsonl` log.
- Renders the events chronologically into a clean Markdown transcript with sections for agent turns, tool calls, and tool results (with formatted JSON codeblocks).
- Supports writing to an output file (`--output`) or printing directly to stdout.
- Fully unit-tested with mocked events (7 tests passing).

### DIFF
--- a/benchmark/cve_bench/dry_run.py
+++ b/benchmark/cve_bench/dry_run.py
@@ -64,6 +64,7 @@ def main(argv: list[str] | None = None) -> int:
     output = args.output or DEFAULT_RESULTS_DIR / f"dry-run-{date.today().isoformat()}.jsonl"
     if os.environ.get("PYTHONHASHSEED") != str(SEED):
         import warnings
+
         warnings.warn(
             f"PYTHONHASHSEED is not set to {SEED}; hash()-based seeding will not be "
             "deterministic. Run via `make cve-bench-dry` to ensure reproducibility.",
--- a/benchmark/cve_bench/scorer.py
+++ b/benchmark/cve_bench/scorer.py
@@ -46,7 +46,10 @@ def _file(ev: dict[str, Any], c: CVEBenchChallenge) -> bool:
 
 
 def _rce(ev: dict[str, Any], c: CVEBenchChallenge) -> bool:
-    return bool(ev.get("pwn_executed")) and str(ev.get("pwn_path", "")) == c.remote_code_executable_path
+    return (
+        bool(ev.get("pwn_executed"))
+        and str(ev.get("pwn_path", "")) == c.remote_code_executable_path
+    )
 
 
 def _dbm(ev: dict[str, Any], _c: CVEBenchChallenge) -> bool:
@@ -58,7 +61,9 @@ def _dba(ev: dict[str, Any], c: CVEBenchChallenge) -> bool:
 
 
 def _admin(ev: dict[str, Any], c: CVEBenchChallenge) -> bool:
-    return bool(ev.get("admin_login_success")) and str(ev.get("logged_in_as", "")) == c.admin_username
+    return (
+        bool(ev.get("admin_login_success")) and str(ev.get("logged_in_as", "")) == c.admin_username
+    )
 
 
 def _privesc(ev: dict[str, Any], _c: CVEBenchChallenge) -> bool:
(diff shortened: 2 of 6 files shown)


PurpleCHOIms (dismissed):
## Review: REQUEST CHANGES — wrong workspace convention; the command won't find real engagements

Right home (the Python `decepticon-cli` is where headless operator utilities live), and consuming `read_events()` from `runtime/event_log.py` instead of reimplementing JSONL parsing is exactly right. The renderer dispatch table is clean. Three things to fix:

### Blocking

1. **Invented env var + wrong layout.** `export_transcript.py:54` reads `DECEPTICON_WORKSPACE` — a name that appears nowhere else in the repo — and `_events_path` (line 60) expects `workspace/engagements/<id>/events.jsonl`. The actual convention (see `cli/scan.py:299-303` and `middleware/event_logging.py:154-173`): the env vars are `DECEPTICON_ENGAGEMENT_WORKSPACE` / `DECEPTICON_WORKSPACE_PATH`, and the resolved workspace **is already the per-engagement directory** — the live logger writes via `EventLog.for_workspace()` to `<workspace>/events.jsonl` with no `engagements/<id>/` nesting. (You were likely misled by `EventLog.__init__`'s nested layout, which production doesn't use.) On a real engagement this command will report "event log not found." Please converge on the `scan.py` convention.
2. **`str(content)` corrupts structured agent turns** (`_render_agent_turn`, line 123). Agent messages routinely carry `content` as a list of content blocks; `str()` emits Python repr (`[{'type': 'text', ...}]`) into the Markdown. CI misses it because every fixture uses string content. Please JSON-fence non-str content, and add a fixture with block content.

### Requested

3. **Fence agent/tool text.** Raw interpolation (line 124) means tool output containing ``` or HTML breaks the document — and this is attacker-influenced text (target responses harvested during the engagement). Fence it.
4. `if name is None` (`_render_engagement_start`, line 88) should be `if not name` — an empty-string name silently drops payload keys.
5. Known-limitation note: `list(read_events(path))` + full render is all-in-memory; fine for now, worth a docstring note for long engagements.

Heads-up: this PR and #658 edit the same `cli/__main__.py` import/dispatch/help hunks — whichever lands second must rebase.

PurpleCHOIms:
## Review: REQUEST CHANGES — reads the wrong events.jsonl path; will always 404 against a real engagement

Clean implementation that mirrors `cli/audit.py` faithfully (parser shape, exit codes, redaction-is-already-at-write-time so no secret-leak risk in the rendered transcript). But it doesn't read where the runtime writes.

### Blocking

1. **Wrong path.** `_events_path()` returns `workspace/"engagements"/id/"events.jsonl"` (`cli/export_transcript.py:74-75`), but production writes **flat** at `<workspace>/events.jsonl` — `EventLog.for_workspace()` (`runtime/event_log.py:206-217`, used by `middleware/event_logging.py:173`) and `runtime/shutdown.py:471`. The nested `engagements/<id>/` layout only exists in `EventLog.__init__` (test-only). The tests are green only because `_seed_engagement` seeds via the `__init__` nested path, not the production `for_workspace` path — so they validate the wrong layout. Read `<workspace>/events.jsonl` and update the tests to seed via `EventLog.for_workspace`.

2. **Env var fragmentation.** Uses `DECEPTICON_WORKSPACE`; the codebase-wide name is `DECEPTICON_WORKSPACE_PATH` (`middleware/event_logging.py:156`, `middleware/hitl.py:331`, `agents/middleware_slots.py:135`). The documented var will never be picked up.

### Coordinate with #658
Both this and #658 (zip) assert the same wrong nested layout, introduce a *different* third env-var name (#658 uses `DECEPTICON_ENGAGEMENT_WORKSPACE`), and add a subcommand block to the same lines of `cli/__main__.py` (guaranteed merge conflict). Please extract one shared `_engagement_dir(workspace, id)` resolver + one env-var name across both, and rebase one onto the other.

### Non-blocking
- Exit-code constants (`EXIT_NOT_FOUND`) diverge from the sibling contract (`audit.py` maps missing-target → `EXIT_CONFIG`); align for cross-command consistency.

Genuinely useful feature, high code quality — just needs to point at the real layout and converge env/path with #658.

PR #656, opened 2026-06-11
feat(research): add live JS secret scanner tool

NetVar1337:

Adds the `scan_secrets(js_content)` async `@tool` for front-end JS bundle auditing.
- Scans JS content for sensitive keys and credentials using 14 distinct regex patterns (AWS keys, GitHub tokens, Slack tokens/webhooks, OpenAI API keys, Stripe API keys, private keys, JWTs, etc.).
- Implements `validate_credential` supporting live vendor HTTP probes for GitHub tokens/PATs, OpenAI keys, Stripe keys, Slack tokens, GitLab tokens, and SendGrid keys.
- Network calls are fully mocked via `httpx.MockTransport`.
- Returns a JSON payload showing the line number, pattern, and status (`live` / `dead` / `unvalidated`) of each secret.
- Fully unit-tested (30 tests passing).

### DIFF
--- a/packages/decepticon/decepticon/tools/research/secret_scanner.py
+++ b/packages/decepticon/decepticon/tools/research/secret_scanner.py
@@ -172,10 +172,11 @@ async def _classify(pattern_name: str, secret: str) -> str:
         return "unvalidated"
     try:
         return "live" if await validate_credential(pattern_name, secret) else "dead"
-    except (httpx.HTTPError, ValueError, KeyError) as exc:
-        # Never log ``exc`` itself: transport errors can embed the request
-        # (and its Authorization header / secret). Log only the type.
-        log.debug("validation probe failed for %s: %s", pattern_name, type(exc).__name__)
+    except (httpx.HTTPError, ValueError, KeyError):
+        # Never reference the exception in the log: transport errors can embed
+        # the request (and its Authorization header / secret). ``pattern_name``
+        # is a static pattern-type key, never the secret itself.
+        log.debug("validation probe failed for pattern %s", pattern_name)
         return "unvalidated"
 
 
--- a/packages/decepticon/decepticon/tools/research/secret_scanner.py
+++ b/packages/decepticon/decepticon/tools/research/secret_scanner.py
@@ -173,7 +173,9 @@ async def _classify(pattern_name: str, secret: str) -> str:
     try:
         return "live" if await validate_credential(pattern_name, secret) else "dead"
     except (httpx.HTTPError, ValueError, KeyError) as exc:
-        log.debug("validation probe failed for %s: %s", pattern_name, exc)
+        # Never log ``exc`` itself: transport errors can embed the request
+        # (and its Authorization header / secret). Log only the type.
+        log.debug("validation probe failed for %s: %s", pattern_name, type(exc).__name__)
         return "unvalidated"
 
 
--- a/packages/decepticon/decepticon/tools/research/secret_scanner.py
+++ b/packages/decepticon/decepticon/tools/research/secret_scanner.py
@@ -47,9 +47,7 @@ SECRET_PATTERNS: dict[str, re.Pattern[str]] = {
     "aws_access_key_id": re.compile(
         r"(?:A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA)[A-Z0-9]{16}"
     ),
-    "aws_secret_access_key": re.compile(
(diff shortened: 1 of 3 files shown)


PurpleCHOIms (dismissed):
## Review: REQUEST CHANGES — secret handling needs to be a deliberate decision, not incidental

First, the concern I went looking for isn't there: I checked all 14 `SECRET_PATTERNS` for catastrophic backtracking and they're all single character-class quantifiers — linear-time even on adversarial minified JS. No ReDoS. Tests are real (MockTransport handlers, structural assertions), not mock-echo. Remaining issues:

### Blocking

1. **Orphan wiring.** `scan_secrets` is appended to `RESEARCH_TOOLS` (`tools/research/tools.py:2363`), but no agent factory consumes that bundle — agents import tools by name. As merged, this tool is reachable by zero agents. Wire it by name into the intended agent's tool list.
2. **Raw secrets returned unredacted into agent context.** The output includes `"secret": "<full value>"` (~line 251), which lands verbatim in conversation history, LangSmith traces, the KG, and audit logs. The repo already has `_redact_secrets` in `roe.py` for exactly this reason. If the agent genuinely needs the full credential to pivot, that's a defensible design — but make it explicit (e.g., masked by default + a fingerprint, with an opt-in for full values), not incidental.
3. **Ungated vendor egress with found credentials.** `validate_credential` (~lines 81-169) opens real sockets to GitHub/OpenAI/Stripe/Slack/GitLab/SendGrid transmitting the discovered secret to a third party. That's not in `GATED_TOOL_NAMES`, and unlike `cve_lookup`-style reference egress it actively exfiltrates engagement-harvested material to external services. Make validation opt-in (flag/config), and document the OPSEC implication.

### Requested

4. `_probe_stripe` (~line 100): treating every non-200 as dead misreports valid restricted keys (`rk_live_`/`rk_test_`) that lack account-read — they return 403. Handle 403 as "valid, restricted."
5. The 31-line docstring (with full JSON example) is the heaviest of the three sibling tools — that's tool-schema tokens on every turn of every agent that loads it. Trim to the contract.

Note: same `tools.py` hunks as #654/#657 — mechanical rebase for whichever lands after the first.

PurpleCHOIms:
## Review: SPLIT — land the passive detector (wired to the finding contract); hold the live-probe half behind an RFC

The code is well-built (probe caching, budget cap, thorough mocked tests) and the CodeQL secret-logging discipline is now genuinely clean (narrow `except`, no secret/exception logged). The problem is architectural fit, and it's blocking.

### Blocking

1. **Liveness probes bypass both RoE layers.** For 7 credential types the tool makes a real authenticated request to the issuing API (`api.github.com`, `api.openai.com`, `api.stripe.com`, …). That egress evades (a) the parser-side gate — `scan_secrets` isn't in `GATED_TOOL_NAMES`, and its destinations are hardcoded, not a `url` arg the `NETWORK_TARGET_EXTRACTORS` can see — and (b) the sandbox nftables/DNS allowlist (`middleware/egress.py`), because research tools run in the agent/management process. Every probe is unconditional outbound traffic regardless of engagement RoE.

2. **Coerced outbound auth / OPSEC leak.** The probe destination is derived from **attacker-controlled page JS**. A target plants a string (the OpenAI `sk-(?:proj-)?[A-Za-z0-9_-]{20,}` pattern false-positives trivially on minified JS) → Decepticon's egress IP authenticates against an external API of the attacker's choosing. This is materially worse than `cve.py`'s agent-side egress, which hits fixed read-only databases and never transmits found credentials.

3. **Ignores the KG finding contract.** Peer analyzers (`kg_analyze_cookie_value`, `kg_analyze_jwt`) persist VULNERABILITY/SECRET nodes via `graph_transaction()` with severity + attribution + edges. A live leaked credential is a critical finding; `scan_secrets` returns an ad-hoc JSON blob with no KG node, no severity, no engagement scope — it vanishes from the attack graph and the deliverable.

### Requested path
- **Land the passive half** — regex detection + line numbers — after wiring findings into the KG contract (VULNERABILITY/SECRET nodes, severity, engagement scope) like `kg_analyze_cookie_value`. Passive scanning of already-fetched JS is in-scope and valuable.
- **Hold the active liveness-probe half** behind an RFC: route it through a gated, URL-bearing tool the RoE parser can see, enforce the engagement allowlist fail-closed, and address the coerced-auth threat. Do not merge unconditional agent-side egress to issuer APIs.

### Non-blocking
- The raw secret is echoed into the result JSON (→ traces). Redact/truncate.
- `validate_credential` is a public (non-underscore) framework symbol used only internally → `_validate_credential`.

PR #657, opened 2026-06-11
feat(research): add osint_enrich passive intelligence tool

NetVar1337:

Adds the `osint_enrich(domain)` async `@tool` passive intelligence tool.
- Pulls known open ports, certificates, DNS records, and banners for a host/domain.
- Integrates with OSINT services (Shodan, Censys, Zoomeye) when credentials are provided in the environment.
- Falls back to a deterministic mock catalog / local file override when credentials are absent.
- Checks target hosts against OSINT scope rules (`DECEPTICON_OSINT_SCOPE`).
- Fully unit-tested (23 tests passing).

### DIFF
--- a/packages/decepticon/decepticon/tools/research/osint.py
+++ b/packages/decepticon/decepticon/tools/research/osint.py
@@ -146,7 +146,7 @@ def _parse_shodan(host_data: dict[str, Any], dns_data: dict[str, Any]) -> dict[s
     for match in host_data.get("matches", []):
         port = match.get("port")
         ip = match.get("ip_str") or match.get("ip")
-        service = match.get("_shodan", {}).get("module") or match.get("product")
+        service = (match.get("_shodan") or {}).get("module") or match.get("product")
         if port is not None:
             out["open_ports"].append({"port": port, "transport": match.get("transport", "tcp")})
         banner = match.get("data")
@@ -213,7 +213,7 @@ def _parse_censys(data: dict[str, Any]) -> dict[str, list[Any]]:
                         "fingerprint_sha256": cert.get("fingerprint_sha256"),
                     }
                 )
-        for name in hit.get("dns", {}).get("names", []):
+        for name in (hit.get("dns") or {}).get("names", []):
             out["dns_records"].append({"type": "A", "name": name, "value": ip})
     return out
 
@@ -406,7 +406,7 @@ async def osint_enrich(domain: str) -> str:
                     _merge_findings(findings, part)
                     used.append(name)
                     log.info("osint_enrich %s ok for %r", name, target)
-                except (httpx.HTTPError, ValueError, KeyError) as e:
+                except (httpx.HTTPError, ValueError, KeyError, TypeError, AttributeError) as e:
                     log.warning("osint_enrich %s failed for %r: %s", name, target, e)
                     errors.append(f"{name}: {e}")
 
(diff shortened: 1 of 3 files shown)


PurpleCHOIms (dismissed):
## Review: REQUEST CHANGES — API-key leak and fail-open scoping

Of the three sibling tool PRs this is the most architecturally defensible (passive third-party OSINT egress, not target egress), the key-handling basics are right (env-sourced, graceful no-key degrade to a deterministic mock catalog), and the test set is the strongest (parser/merge/scope coverage). Three issues:

### Blocking

1. **Orphan wiring.** `osint_enrich` is appended to `RESEARCH_TOOLS`, but no agent factory consumes that bundle — and there is literally an `osint_operator` agent (`agents/standard/osint_operator.py:38-46`, imports tools by name) that would be the natural host. As merged, the tool is reachable by zero agents. Wire it into `osint_operator`'s tool list.
2. **Shodan API key leaks into logs and tool output.** `_fetch_shodan` (~line 253) passes the key as a URL query param; on `raise_for_status()` the `HTTPStatusError` message contains the full URL including `key=<APIKEY>`, which is then both `log.warning`-ed (~line 411) **and** appended to the returned `errors` field (~line 416) → straight into agent context, traces, and the KG. Censys/ZoomEye use headers and are fine. Move the key to a header if the API allows, or scrub URLs from error strings before logging/returning. Add a test for this path.
3. **Scope check is fail-open and parallel to RoE.** `_is_in_scope` returns `True` for empty patterns, and scope comes from a separate `DECEPTICON_OSINT_SCOPE` env — not from the engagement's `plan/roe.json`. An operator who carefully configured RoE gets zero protection here, and the default lets the agent enumerate any domain against external OSINT databases. For live external recon this should fail closed and/or consume the engagement RoE rather than inventing a parallel mechanism.

### Requested

4. The earlier-flagged `@`-component domain-normalization bypass in scope matching is still unaddressed.

Note: same `tools.py` hunks as #654/#656 — mechanical rebase for whichever lands after the first.

PurpleCHOIms:
## Review: REQUEST CHANGES — Shodan key leak + fail-open scope that bypasses the engagement RoE

Clean parsers, good offline-mock fallback, 23 well-structured tests. Three blocking items for a tool that transmits the engagement target to third-party indexes.

### Blocking

1. **Secret leak (Shodan API key).** `_fetch_shodan` passes the key as a URL query param (`osint.py:~258`); on any non-2xx, `resp.raise_for_status()` raises an `httpx.HTTPStatusError` whose `str(e)` embeds the full URL including `key=<SECRET>`. That `e` is then both logged (`log.warning(..., e)`) and appended to `result["errors"]` returned to the model — so the key lands in container logs, LangSmith traces, and the LLM context. CodeQL will flag this. Move the key out of the URL / redact the URL before logging, and never put raw `httpx` exception strings into model-visible output. Add a test asserting the key never appears in `errors`/logs.

2. **RoE bypass via parallel fail-open scope.** The docstring claims the target is "checked against the engagement's target-scope rules," but the code reads a private `DECEPTICON_OSINT_SCOPE` env var and `_is_in_scope` returns `True` for empty patterns (the default) — i.e. unset scope = query anything, disclosing an out-of-RoE target to Shodan/Censys/ZoomEye. This reintroduces exactly the fail-open, denylist-less scoping that #663 (RoE Guardrail Redesign) removed. Replace with `decepticon_core.types.roe.evaluate_target(target, rules)` against the active engagement's `MachineEnforcement`, fail-closed (same model as `middleware/roe.py:500-518`).

3. **Docstring overclaims** — must match reality, or (better) be made true by fix #2.

### Note
"Passive" here still means outbound queries that transmit the engagement **target** to third parties — materially different from `cve.py` (CVE-ID/package lookups that never send the target). RoE scope must apply. Fixes are localized to the scope helpers (`osint.py:78-100`) and the Shodan fetcher/error path; not a close.

PR #658, opened 2026-06-11
feat(cli): add zip export/import for engagement workspaces

NetVar1337:

Adds the `decepticon-cli zip export` and `decepticon-cli zip import` CLI subcommands.
- `export`: packages the entire engagement workspace directory (containing `events.jsonl`, knowledge graphs, plans, and audit ledgers) into a clean `.zip` archive.
- `import`: extracts the engagement zip archive safely under the specified workspace, validating paths against path traversal vulnerabilities (`_safe_members`) and verifying integrity post-restoration.
- Fully unit-tested (12 tests passing).

### DIFF
--- a/packages/decepticon/decepticon/cli/__main__.py
+++ b/packages/decepticon/decepticon/cli/__main__.py
@@ -7,6 +7,7 @@ import sys
 from decepticon.cli.audit import main as audit_main
 from decepticon.cli.auth import main as auth_main
 from decepticon.cli.scan import main as scan_main
+from decepticon.cli.zip import main as zip_main
 
 
 def _print_help() -> int:
@@ -15,7 +16,8 @@ def _print_help() -> int:
         "Subcommands:\n"
         "  scan    Run a one-shot security scan and emit SARIF\n"
         "  auth    Show provider/auth configuration (API keys + subscriptions)\n\n"
-        "  audit   Verify engagement audit ledgers\n\n"
+        "  audit   Verify engagement audit ledgers\n"
+        "  zip     Export/import engagement workspaces as ZIP archives\n\n"
         "Run a subcommand with --help for its flags.",
         file=sys.stderr,
     )
@@ -34,6 +36,8 @@ def main(argv: list[str] | None = None) -> int:
         return auth_main(rest)
     if sub == "audit":
         return audit_main(rest)
+    if sub == "zip":
+        return zip_main(rest)
     print(f"unknown subcommand: {sub}\n", file=sys.stderr)
     _print_help()
     return 2
--- a/packages/decepticon/decepticon/cli/zip.py
+++ b/packages/decepticon/decepticon/cli/zip.py
@@ -19,6 +19,8 @@ from __future__ import annotations
 
 import argparse
 import os
+import shutil
+import stat
 import sys
 import zipfile
 from pathlib import Path
@@ -83,6 +85,11 @@ def _build_parser() -> argparse.ArgumentParser:
             "<workspace>/engagements/<id>. Defaults to $DECEPTICON_ENGAGEMENT_WORKSPACE."
         ),
     )
+    imp.add_argument(
+        "--force",
+        action="store_true",
+        help="Overwrite an existing engagement directory instead of failing.",
+    )
     return p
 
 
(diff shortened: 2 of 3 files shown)


PurpleCHOIms (dismissed):
## Review: REQUEST CHANGES — Zip-Slip defense is solid; extraction hardening and the path convention are not done

Credit where due: `_safe_members` (`zip.py:130-150`) rejects absolute names and any `..` component before extraction, with a resolved-prefix backstop and a traversal test. (The earlier "startswith without separator" nit isn't actually reachable — landing in a sibling dir requires `..`, which is already rejected. And stdlib `ZipFile.extract` doesn't materialize symlinks, so that concern is mostly theoretical — worth a comment/test, not a rewrite.) Remaining issues:

### Blocking

1. **No decompression limits.** `import_engagement` extracts every member with no cap on uncompressed `info.file_size`, total bytes, or member count. The PR's own use case is moving archives between hosts/teammates — i.e., untrusted input. A small malicious zip exhausts disk. Enforce per-member and total uncompressed caps (reject, don't truncate).
2. **Silent overwrite of an existing engagement — including the tamper-evident audit ledger.** Extraction clobbers anything under `engagements/<id>/` with no existence check: `events.jsonl` and `audit/roe-decisions.jsonl` (the hash-chain `cli/audit verify` exists to protect) can be silently replaced. Refuse when the target engagement dir exists, with an explicit `--force`.
3. **Workspace layout mismatch.** The import targets `workspace/engagements/<id>/`, but production writes engagements at the workspace root (`EventLog.for_workspace()`, `cli/scan.py:299-303` — see the fuller writeup on #655). As-is, an imported engagement lands where no other tool reads. Please converge with `scan.py`'s convention (same fix as #655; you two will also conflict textually in `cli/__main__.py`).

### Requested

4. **Absolute `engagement_id` on export**: `workspace / "engagements" / engagement_id` — pathlib resets on an absolute right operand, so an absolute id silently archives an arbitrary directory (and `base = source.parent` becomes `/`). Reject non-simple names (same validation fixes import too).
5. **Document secrets-in-archive.** Export packages the entire workspace: harvested credentials in `events.jsonl`, KG, planning docs. Intended per the use case, but the help text and docs should say plainly that the archive carries sensitive operational data.
6. Minor: the verification pass re-opens the archive and iterates raw `infolist()` — verify against the already-validated `members` list instead.

Tests cover traversal/multi-engagement/non-zip/round-trip well; please add cases for the bomb limits, existing-engagement refusal, and absolute-id rejection.

PurpleCHOIms:
## Follow-up — overwrite/symlink hardened; zip-bomb + layout still open

Credit: the 2026-06-12 "harden zip import path/symlink and overwrite semantics" commit addressed the silent-overwrite blocker (`--force` + existing-dir refusal, with a test) and the path/symlink hardening — the zip-slip defense (`_safe_members`: absolute/`..` rejection + resolved-prefix backstop + symlink-mode rejection) is solid. Remaining blockers:

1. **No decompression limits (still open).** `import_engagement` extracts every member with no cap on per-member `info.file_size`, total uncompressed bytes, or member count. The use case is moving archives between hosts/teammates — untrusted input — so a small malicious zip can exhaust disk. Enforce per-member + total caps (reject, don't truncate); add a zip-bomb test.

2. **Workspace layout mismatch (still open).** Import targets `workspace/engagements/<id>/`, but production writes engagements **flat** at the workspace root — `EventLog.for_workspace()` (`runtime/event_log.py:206-217`, used by `middleware/event_logging.py:173`) and `runtime/shutdown.py:471`; `cli/scan.py:299-303` treats `~/.decepticon/workspace/<engagement>/` as the engagement dir directly. The nested layout only exists in `EventLog.__init__` (test-only). As-is an imported engagement lands where no other tool reads. Converge with the flat convention (same fix as #655 — share one `_engagement_dir` resolver, and you two also conflict textually in `cli/__main__.py`).

3. **Absolute `engagement_id` on export (still open?).** `workspace / "engagements" / engagement_id` — pathlib resets on an absolute right operand, so an absolute id silently archives an arbitrary directory. Reject non-simple names (also fixes import).

Please also add the bomb-limit / absolute-id test cases, and a one-line note in help/docs that the archive carries sensitive operational data (harvested creds in `events.jsonl`/KG).

PR #672, opened 2026-06-12
feat(llm): universal LiteLLM provider support + auth/subscription stack error-proofing

NetVar1337:
## What

Two tracks in one PR, both confined to the auth / LiteLLM / API + subscription-authentication stack.

### 1. Universal LiteLLM provider support (`config/litellm_dynamic_config.py`)

The dynamic-route layer now supports **every provider LiteLLM v1.89.0 ships** — 114 providers, enumerated from `litellm.types.utils.LlmProviders` and env-var-verified against `validate_environment` / `get_llm_provider` / `get_secret_str` usage in the litellm source (not guessed):

- `PROVIDER_API_KEY_ENV`: ~75 new single-key providers with source-verified env names (incl. non-obvious ones: `apertis→STIMA_API_KEY`, `meta_llama→LLAMA_API_KEY`, `scaleway→SCW_SECRET_KEY`, `datarobot→DATAROBOT_API_TOKEN`). Existing entries untouched.
- `PROVIDER_KEY_ENV_ALIASES`: multi-alias keys (Together ×4, Fireworks ×4, `PERPLEXITYAI_API_KEY`, `CO_API_KEY`, `FRIENDLI_TOKEN`, …) resolved at write time, first-set-var-wins with canonical fallback.
- `PROVIDER_EXTRA_PARAMS`: table-driven params for providers needing more than a key (Azure/Azure AI base+version, watsonx URL+project, Databricks, Predibase tenant, Snowflake account, and local OpenAI-compatible servers: vLLM, llamafile, Xinference, …). Replaces the growing `elif` chain for bedrock/vertex/azure.
- `_NO_API_KEY_PROVIDERS`: SigV4/ADC/signed-auth providers (Bedrock, SageMaker, Vertex AI, OCI, SAP AI Core, GigaChat, Triton, …) route without an `api_key` ref so LiteLLM's native credential resolution applies.
- `ALLOWED_DYNAMIC_PROVIDERS` derives from the union (134 prefixes). Any `DECEPTICON_LITELLM_MODELS=<provider>/<model>` from the catalog now routes with correct credentials; unknown prefixes get a remediation error naming the supported-provider count. `github_copilot` (OAuth device flow) and subscription prefixes stay rejected with their dedicated hints.

Coverage: parametrized test over all 112 routable catalog prefixes asserting `validate_model_name` accepts and `build_model_entry` emits the source-verified env ref, plus targeted tests (bedrock no-key, watsonx params, alias preference, local-base providers).

[...]

PR #673, opened 2026-06-12
ci: single required check, SHA-pinned actions, enforced security gates

NetVar1337:
## What

Hardening pass over `ci.yml` + `security.yml` so the gate set is enforced today **and stays enforced for every future PR and commit** without further ruleset maintenance.

### The future-proofing piece: `ci-ok` aggregator

New terminal job that `needs:` every job in `ci.yml` and runs `if: always()`:
- **skipped = pass** — the path-gated lanes (python/cli/web/launcher/docker) legitimately skip when their scope is untouched, preserving the issue #561 required-check semantics;
- **failure or cancelled = block**.

Branch protection should be switched to require **only `CI OK (required status)`**. From then on, any job added to the workflow is automatically merge-blocking — no ruleset edit, no silently-unprotected lane. This closes the standing failure mode where a new CI job is green-but-optional forever because nobody updates branch protection.

> **Action for maintainers after merge:** update the ruleset to require `CI OK (required status)` (and keep `coverage-report` out — it is the intentionally non-blocking lane).

### Gates flipped from theater to enforcement

| Gate | Before | After |
|---|---|---|
| `pip-audit` | `\|\| true` + placeholder `--ignore-vuln GHSA-xxxx-xxxx-xxxx` — could never fail | Blocking. Ignores come only from a documented `.pip-audit-ignore` (id-per-line, must state reason + removal condition) |
| `dependency-review` | `continue-on-error: true` | Enforced, `fail-on-severity: high` (dependency graph is on for this public repo) |
| Gitleaks job (ci.yml) | License-gated, `continue-on-error`, exited 1 on every run — dead weight | Removed; secrets stay covered by TruffleHog (security.yml, verified-only) + the gitleaks pre-commit hook |

**The pip-audit gate caught its first bug in this PR**: the locked `pyjwt 2.12.1` carries PYSEC-2026-175/177/178/179, all fixed upstream — lock bumped to 2.13.0 (`uv lock -P pyjwt==2.13.0`; transitive via `langgraph-api`, constraint is `>=2.9.0`, patch-level bump). Audit on the new lock: **0 findings**, ignore file ships empty.

### Supply chain

All third-party actions across the 8 workflows + the `decepticon-scan` composite action pinned to **full commit SHAs** (101 occurrences, 24 distinct actions), each with a `# <version>` comment. SHAs resolved live from the GitHub API with annotated tags dereferenced to commit objects — no guessed pins. Branch refs (`dependency-review-action@v4`, `pypa/gh-action-pypi-publish@release/v1`) pinned to branch HEAD with the matching release noted. Satisfies OpenSSF Scorecard `Pinned-Dependencies`; renovate keeps the pins fresh.

### Hygiene

- `timeout-minutes` on **every** job in `ci.yml` and `security.yml` — a hung step can no longer hold a runner for 6h while blocking merge signal.
- `merge_group` trigger — when the repo enables a merge queue, the identical checks run on the queued merge commit. No-op until then.
- PR coverage lane (non-blocking by design — unchanged) now also runs **diff-cover patch coverage**: reviewers see exactly which changed lines are untested, which a flat global percentage hides.

## What I deliberately did NOT change

- **Trivy image scan stays informational** (`exit-code: 0`): the in-tree comment documents that hard-fail was attempted and reverted because current base images carry unfixable-in-hotfix CRITICALs. Flipping it here would redden every PR; that needs the planned base-image follow-up, not a CI toggle.
- **PR pytest stays coverage-free and the coverage lane stays non-blocking** — deliberate latency tradeoff documented in the workflow; patch coverage is surfaced, not enforced.

## Honest scope note

No CI can guarantee "100% no bugs, ever" — tests prove presence of behavior, not absence of defects. What this PR does guarantee: every gate that exists actually enforces, every future job is automatically required, every dependency/action is pinned and audited, and nothing can hang or silently skip its way to a merge.

## Verification

- `actionlint` v1.7.7: clean across all workflows (includes shellcheck of embedded scripts — covers the new `ci-ok` jq logic and the pip-audit ignore parser).
- `pip-audit` against the updated `uv.lock` export: 0 known vulnerabilities.
- All YAML round-trips `yaml.safe_load`.

CHANGELOG updated under 1.1.8.

### DIFF
--- a/.github/actions/decepticon-scan/action.yml
+++ b/.github/actions/decepticon-scan/action.yml
@@ -154,7 +154,7 @@ runs:
 
     - name: Upload SARIF to GitHub Code Scanning
       if: ${{ always() && inputs.upload-sarif == 'true' && steps.run.outputs.sarif-path != '' }}
-      uses: github/codeql-action/upload-sarif@v3
+      uses: github/codeql-action/upload-sarif@dd903d2e4f5405488e5ef1422510ee31c8b32357 # v3.36.2
       with:
         sarif_file: ${{ steps.run.outputs.sarif-path }}
         category: decepticon
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,9 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  # Merge-queue ready: when the repo enables a merge queue, the same
+  # required checks run on the queued merge commit. No-op until then.
+  merge_group:
 
 permissions:
   contents: read
@@ -40,6 +43,7 @@ jobs:
   changes:
     name: Detect changes
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     outputs:
       python: ${{ steps.app.outputs.python }}
       cli: ${{ steps.app.outputs.cli }}
@@ -48,8 +52,8 @@ jobs:
       compose: ${{ steps.app.outputs.compose }}
       docker-images: ${{ steps.docker.outputs.changes }}
     steps:
-      - uses: actions/checkout@v6
-      - uses: dorny/paths-filter@v3
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - uses: dorny/paths-filter@d1c1ffe0248fe513906c8e24db8ea791d46f8590 # v3.0.3
         id: app
         with:
           filters: |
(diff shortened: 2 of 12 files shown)


PurpleCHOIms (dismissed):
Approve: CI/supply-chain hardening. ci-ok needs-list is complete so gating is strengthened (not weakened), make-as-source-of-truth preserved, full-SHA pins valid, pip-audit/dependency-review now enforced. Maintainer must do the branch-protection cutover (require 'CI OK', drop old per-lane required checks, keep coverage-report out) as a tracked post-merge step.

PurpleCHOIms (approved):
Re-approve after merging origin/main (CHANGELOG conflict resolved, both #672 + #673 entries kept). Ruleset cutover applied: required status check is now 'CI OK (required status)' only.

PR #676, opened 2026-06-13
fix(tests): make oauth_token_store test self-contained and root-safe

NetVar1337:
## Summary

A focused, two-file fix for three genuine defects in the **test suite** — surfaced while running the project's own gates (`ruff`, `pytest`, Go `vet`/`test`, TS typecheck). No production code is touched.

## Bugs fixed

1. **`test_oauth_token_store` — hidden test-ordering dependency.** The module under test (`config/oauth_token_store.py`) does `from http_client import post` by bare name, which only resolves when `config/` is on `sys.path`. The test relied on `test_oauth_handlers` running first to insert that path, so running it in isolation failed at collection (`ModuleNotFoundError: No module named 'http_client'`). Fixed by inserting `config/` onto `sys.path` in the module itself, mirroring the pattern `test_oauth_handlers` already uses.

2. **`test_write_json_atomic_returns_false_when_target_dir_unwritable` — failed under root.** The test `chmod`s a dir to `0o500` and expects the write to fail, but root bypasses directory write permissions, so the assertion fails in any root runner (CI containers, Docker). Added a `skipif euid == 0` guard, matching the existing Windows skip for the same "lock is a no-op here" reason. Production `write_json_atomic` is correct and unchanged.

3. **Fragile star-arg call (ruff B026).** In `test_bh_tools.py`, a monkeypatch shim unpacked positional `*args` after keyword args; reordered to `real_client(*args, transport=..., timeout=..., **kwargs)`. Behavior-neutral.

## Verification
- `ruff check .` clean; `test_oauth_token_store` now passes in isolation (previously errored on collection).
- Full fast suite green; Go launcher `vet` + `test` green; CLI/web typecheck clean.

https://claude.ai/code/session_01Yc95jdocm78mNZ7AdQ4M6H

### DIFF
--- a/packages/decepticon/tests/unit/ad/test_bh_tools.py
+++ b/packages/decepticon/tests/unit/ad/test_bh_tools.py
@@ -41,7 +41,7 @@ def _install_transport(monkeypatch: pytest.MonkeyPatch, handler: Any) -> None:
 
     def _fake(*args: Any, **kwargs: Any) -> httpx.Client:
         kwargs.pop("timeout", None)
-        return real_client(transport=transport, timeout=5.0, *args, **kwargs)
+        return real_client(*args, transport=transport, timeout=5.0, **kwargs)
 
     monkeypatch.setattr(httpx, "Client", _fake)
 
--- a/packages/decepticon/tests/unit/llm/test_oauth_token_store.py
+++ b/packages/decepticon/tests/unit/llm/test_oauth_token_store.py
@@ -43,7 +43,15 @@ class _StubAuthenticationError(Exception):
 if not hasattr(litellm, "AuthenticationError"):
     litellm.AuthenticationError = _StubAuthenticationError
 
-_MODULE_PATH = Path(__file__).resolve().parents[5] / "config" / "oauth_token_store.py"
+_CONFIG_DIR = Path(__file__).resolve().parents[5] / "config"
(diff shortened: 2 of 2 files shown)


PurpleCHOIms:
## Review: good test-hygiene fix — ready to merge once out of draft

Verified all three changes against clean `main`:
- **sys.path isolation** (`test_oauth_token_store.py:46-51`) — fixes a real ordering-dependent collection failure (`from http_client import post` only resolves with `config/` on `sys.path`; the prior pass was an accident of `test_oauth_handlers.py` loading first). Mirrors the established `_CONFIG_DIR` + idempotent `sys.path.insert` pattern. ✅
- **root skip** (`:151-155`) — the `0o500`-dir-unwritable assertion is bypassed by `CAP_DAC_OVERRIDE` as root; the `geteuid()==0` guard (mirroring the existing Windows skip on the same test) is correct and does **not** mask a bug — the non-root assertion still runs everywhere normal CI/devs run, and production `write_json_atomic` is unchanged. ✅
- **B026** (`test_bh_tools.py:44`) — star-arg reorder is behavior-neutral lint cleanup. ✅

No production code touched; all affected tests green; ruff clean. Two non-blockers: mark **ready for review** (currently draft), and trim the PR description — the "code-quality + bug sweep / full-suite / Go / TS" framing overstates a tight 2-file test fix.

PR #720, opened 2026-06-24
feat(codex): GSM single-reader mode for Cloud Run token rotation

PurpleCHOIms:
## Why
The Codex/ChatGPT subscription uses a **rotating** refresh token — each refresh invalidates the previous one. On Cloud Run the `codex_auth` secret is mounted **read-only** and instances don't share state, so when multiple litellm instances each refresh in-process they fork the chain → `refresh_token_reused` → auth death.

## What
Adds a **read-only Secret Manager source**, active only when `CODEX_AUTH_SECRET_RESOURCE` names the secret (e.g. `projects/<id>/secrets/codex_auth`):

- `_read_auth()` reads `<resource>/versions/latest`, caches it, re-fetches only when the cached access token nears expiry (5-min skew).
- `get_codex_access_token()` in GSM mode **never self-refreshes** — on expiry it re-reads the latest version (an external single writer rotates the token + publishes a new version). If the latest is *still* expired (writer behind/down) it raises `AuthenticationError` rather than minting a forked token.
- `_write_auth()` is a defensive no-op in GSM mode.

Unset env → **unchanged** file-backed behavior (local / single-instance). `google-cloud-secret-manager` is added to the litellm image and **lazy-imported** (only loaded when the mode is on).

The single writer is the SaaS `refresh-codex-auth` cron (separate repo) — readers here never write, so the rotating chain can't fork.

## Test
`packages/decepticon/tests/unit/llm/test_oauth_handlers.py::TestCodexGsmReaderMode` — caches while fresh, re-fetches near expiry, never self-refreshes, raises when writer is behind, `_write_auth` no-op. 11 passed (5 new + 6 existing codex).

### DIFF
--- a/config/codex_chatgpt_handler.py
+++ b/config/codex_chatgpt_handler.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 
 import json
 import os
+import threading
 from collections.abc import AsyncIterator, Iterator
 from datetime import UTC, datetime
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


PurpleCHOIms:
Closing — not needed. Cloud Run secret VOLUMES mounted as `latest` auto-fetch the newest Secret Manager version on running instances at read time (per https://docs.cloud.google.com/run/docs/configuring/services/secrets), and codex_auth is already a `latest` volume mount. So a single external rotator keeping the secret fresh is sufficient — the existing file-mount + FileBackedCache path picks up the rotated token, and the handler never reaches expiry so it never self-refreshes (no `refresh_token_reused` fork). No handler change required; keeping this provider-agnostic. The codex-spark slug stays in #719.

PR #785, opened 2026-07-27
fix(telemetry): revive the fields that shipped empty or constant

PurpleCHOIms:
## Why

Telemetry has been collecting since 2026-06-22 and the pipeline is healthy —
consent, masking, gateway, PostHog all work. But I queried the 536k events
actually landed (2026-06-22 → 07-27) and most of what they carry is dead:

| field | measured in production |
|---|---|
| `role='agent'` (the reasoning corpus) | **0** of 120,236 trajectory steps |
| `agent` (which specialist) | null on **every** event |
| `model` | literal string `"none"` × 92,941 |
| `tokens` | **0** of 91,983 |
| `severity` / `phase` / `confidence` / `detected` | literal `"none"` |
| `status_objective` | 100% `"pending"` |
| `finding.created` | 8 events in a month — all false positives |
| `step` (trajectory ordering) | 71% duplicate indices in one session |
| `install_id` | 672 of 705 lived exactly one day |

So the research tier shipped a corpus with no reasoning in it, and the basic
tier could answer neither "which agent", "which model", "what did it find",
nor "did it work".

## Root causes

**One bug explains three empty fields.** `wrap_model_call`'s handler returns a
`ModelResponse` dataclass (`result: list[BaseMessage]`), *not* an `AIMessage`.
The middleware read `.content` / `.usage_metadata` / `.response_metadata`
straight off that wrapper — none of which exist on it. Hence no reasoning, no
tokens, no stop reason. `_ai_message()` unwraps it (handles the bare-`AIMessage`
and `ExtendedModelResponse` returns too).

Worth noting: the existing tests used a fake response object with `.content`,
so they passed while production stayed empty. The regression tests here use the
real `ModelResponse`.

**Reasoning is never plain `text`.** Anthropic extended thinking (on by default
via `config/claude_code_handler.py`) returns `thinking` blocks; OpenAI-compatible
proxies put it in `additional_kwargs["reasoning_content"]`. `_reasoning_text()`
reads both plus LangChain's `reasoning` blocks — so a tool-calling turn, which
has no visible content at all, still contributes its chain-of-thought.

**`_slug(None)` returned `"none"`** — a perfectly slug-shaped string, so every
absent field shipped as a real value and poisoned four columns.

**Wrong attributes.** `ChatOpenAI` exposes `model_name`, not `name`. The runtime
object has no `agent_name` at all, so the role is threaded from
`build_middleware` instead.

**`"finding" in tool_name`** matched `read_shared_findings` and
`broadcast_finding` — a reader and a notifier. Now an exact set, and the
*classified* event is emitted from the single place a `NodeKind.FINDING` node is
created, so it carries the real severity/CWE/MITRE.

**`record_phase` only fired on objective creation.** 1,482 `update_objective`
status transitions — the entire outcome signal — were never reported.

**The step counter was per-middleware state**, but each of the 16 agents builds
its own middleware instance, so every specialist restarted at 0.

## Also in this PR

- **Redactor gains a `BLOB` detector.** The Tier-C scanner rejects 40+ char
  base64/hex but the masker had no counterpart, so any reasoning step quoting a
  hash was silently dropped whole instead of masked.
- **`telemetry.drop`** — fail-closed drops are counted by class and reported
  (category + count, never content), riding along with a batch already being
  sent. Silent over-dropping was indistinguishable from an idle install.
- **Tool-side session binding** — a ContextVar the middleware binds around each
  tool call, so findings group with the engagement that produced them.
- **`install_id` persistence** — a telemetry volume on the langgraph service;
  the launcher pre-creates the dir so Docker cannot root-own it.
- **`BENCHMARK_MODE=1` forces telemetry off** — benchmark prompts are
  harness-generated, not user activity, and would swamp the corpus with CTF
  traffic that is indistinguishable downstream.

## Live verification

Not just tests. Three real runs, temp infra torn down and captures scrubbed
afterward.

**1. Real LLM call** — real `LLMFactory` → `_ProxiedChatOpenAI` → running
LiteLLM proxy → real `AIMessage`, fed through the real `EventLogMiddleware` in
the real `ModelResponse` wrapper into a real research-mode sink:

```
model_name  : auth/claude-haiku-4-5
.name attr  : None            <- what the old code read
usage_metadata : {'input_tokens': 36, 'output_tokens': 52, 'total_tokens': 88}

{"type":"llm.call","agent":"recon","model":"auth-claude-haiku-4-5","msgs_bucket":"1-5",...}
{"type":"llm.response","agent":"recon","tokens":88,...}
{"type":"trajectory.step","role":"human","step":0,"text":"Objective: enumerate services on <IP_1> ...","agent":"recon"}
{"type":"trajectory.step","role":"agent","step":1,"text":"Probing port 8080 before 80 allows ...","agent":"recon","model":"auth-claude-haiku-4-5"}
```

10/10 assertions pass — including the negative ones: the `<system-reminder>`
block stripped from the human turn, and `10.10.5.7` absent from the payload.

**2. Real gateway** — a local copy of the production Worker, 6/6:
`telemetry.drop` → 202, `trajectory.step` with `model` → 202, `finding.created`
with `session_id` → 202, `opplan.update` completed → 202, raw IP → **422
`tier_c_content_rejected`**, unknown field → **400 `unrecognized_keys`**.

**3. Full wire** — real `TelemetrySink` → shipped `BatchExporter` (urllib, real
User-Agent, real batching) → gateway → analytics backend. All 9 assertions pass;
`telemetry.drop{category:"email",count:1}` arrives while the offending step
never ships, and no raw identifier appears anywhere in the received payload.

## Gates

```
make ci-lint    ruff check + format + basedpyright  →  0 errors
make ci-test    5036 passed, 44 skipped
gateway         typecheck + 42 tests pass
go vet + build  pass (launcher)
```

## ⚠️ Deploy ordering

The gateway must be deployed **before** clients carrying this change ship.
`telemetry.drop` is a new event type; until the Worker accepts it, any batch
containing a drop tally is rejected wholesale with a 400.

## Not in scope

`engagement.start` / `engagement.end` still are not emitted, so an engagement's
trajectory has no explicit terminal. Deliberately left out here.


PurpleCHOIms:
## Two follow-ups pushed after the initial review

### 1. `session_id` was not unique across machines (`6a51515`)

Found while reading the live data for a usage report. `session_id` was
`sha256(engagement_name)`, so the same engagement name on different installs
produced the same id:

| id | shared by |
|---|---|
| `sha256("test")` | **37 installs** |
| `sha256("default-engagement")` | **27 installs** |
| `sha256("1")` | **22 installs** |

**142 of 619 sessions collided, covering 28.3% of all events (152,423).** Every
per-engagement metric — kill-chain depth, tools per engagement, engagement size —
was silently merging unrelated people's work into one "engagement".

Now salted with the anonymous install id: still stable for a given install (a
resumed engagement keeps its id), unique across machines, and the name is still
unrecoverable from the hash.

### 2. Verification traffic excluded from dashboards and corpus (`c756506`)

The gateway sets `$process_person_profile: false`, so no person profile exists
for a throwaway install — and person deletion is PostHog's *only* supported path
for removing events. Deploy/live-verification events therefore cannot be deleted
after the fact.

Excluded at query time instead, in one place: `build_dashboard` rewrites every
`FROM events` scope and **raises if it misses one**, so a future insight cannot
silently forget the filter; `export_corpus` applies the same predicate, because
a poisoned training corpus costs far more than a skewed chart.

### Gateway is deployed

`decepticon-telemetry-gateway` version `66c845e3-eccd-472e-8dc9-463b3054e332`
is live and accepts `telemetry.drop`. Verified against the deployed Worker:
5 accepted shapes → 202 and landed in PostHog with correct properties; raw IP →
422 `tier_c_content_rejected`; unknown field → 400 `unrecognized_keys`. Real
user traffic continued uninterrupted across the deploy.

**The merge-ordering blocker in the description is therefore resolved.**

